PR #13, opened 2021-05-17
refactor:Add missing activity id in camunda event handler

dharmendrak:
### Change description ###

refactor:Add missing activity id in camunda event handler

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/caseevents/ProceedOfflineCallbackHandler.java
```diff
@@ -19,9 +18,8 @@
 import static uk.gov.hmcts.reform.civil.callback.CallbackType.ABOUT_TO_SUBMIT;
 import static uk.gov.hmcts.reform.civil.callback.CaseEvent.PROCEEDS_IN_HERITAGE_SYSTEM;
 
-@Service
 @RequiredArgsConstructor
-public class ProceedOfflineCallbackHandler extends CallbackHandler {
+public abstract class ProceedOfflineCallbackHandler extends CallbackHandler {
```
pd-dudkiewicz, line 22: I think that the only change we need to do in this class is to add `camundaActivityId`. It doesn't really matter what the activity id is until we don't have the same one twice in a row within single business process.

Besides, I don't think that extending this class is going to work as we want it to. Callback handler is selected based on list of events, so if we have the same in more than one handler, then it's going to pick the one that was registered last? That's what I'm assuming by looking at `CallbackHandlerFactory` constructor.

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/docmosis/GenerateDirectionsQuestionnaireCallbackHandler.java
```diff
@@ -28,10 +28,19 @@
...
+    private static final String CLAIMANT_TASK_ID = "ClaimantResponseGenerateDirectionsQuestionnaire";
 
     private final DirectionsQuestionnaireGenerator directionsQuestionnaireGenerator;
     private final ObjectMapper objectMapper;
 
+    @Override
+    public String camundaActivityId(CallbackParams callbackParams) {
+        return callbackParams.getCaseData().getApplicant1ResponseDate() != null
+            ? CLAIMANT_TASK_ID
+            : TASK_ID;
```
pd-dudkiewicz, line 41: As above, I don't think we need that logic, this event is triggered by separate business processes.

pd-dudkiewicz (approved):

PR #92, opened 2021-06-15
fix: Add retries to pull-latest-release-asset.sh (CMC-1688)

adamjhc:

https://tools.hmcts.net/jira/browse/CMC-1688

### Change description ###

Adds retries to pull-latest-release-asset.sh

```
[ ] Yes
[x] No
```

### DIFF
--- a/bin/pull-latest-release-asset.sh
+++ b/bin/pull-latest-release-asset.sh
@@ -4,10 +4,18 @@ repoName=$1
 assetName=$2
 
 retries=0
-while [ -z "$latestRelease" ]
+until [ -f "$assetName" ]
 do
-  latestRelease=$(curl https://api.github.com/repos/hmcts/${repoName}/releases/latest)
+  latestAssetId=$(curl https://api.github.com/repos/hmcts/${repoName}/releases/latest \
+   | docker run --rm --interactive stedolan/jq ".assets[] | select(.name==\"${assetName}\") | .id")
+
+  curl -L \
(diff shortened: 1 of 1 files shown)


bin/pull-latest-release-asset.sh
```diff
@@ -3,8 +3,20 @@
 repoName=$1
 assetName=$2
 
-latestAssetId=$(curl https://api.github.com/repos/hmcts/${repoName}/releases/latest \
-  | docker run --rm --interactive stedolan/jq ".assets[] | select(.name==\"${assetName}\") | .id")
+retries=0
+while [ -z "$latestRelease" ]
```
pd-dudkiewicz, line 7: We should include the second curl call in the retry loop, otherwise the build will fail after failing to pull the asset once. And we could check if `$assetName` file exists as the loop condition.

bin/pull-latest-release-asset.sh
```diff
@@ -3,8 +3,20 @@
...
+do
+  latestRelease=$(curl https://api.github.com/repos/hmcts/${repoName}/releases/latest)
+  retries=$((retries+1))
```
pd-dudkiewicz, line 10: Can we add `echo` call here outputting retry number? Might be useful to know how many attempts it took to pull the asset.

dharmendrak (approved):

pd-dudkiewicz (approved):

swalker125 (approved):

PR #96, opened 2021-06-16
feat:Provide org service address (CMC-1659)

dharmendrak:

CMC-1659

### Change description ###

Provide org service address for claimant and defendant solicitor

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/RepresentativeService.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/RepresentativeService.java
@@ -2,7 +2,10 @@ package uk.gov.hmcts.reform.civil.service.docmosis;
 
 import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.civil.enums.YesOrNo;
+import uk.gov.hmcts.reform.civil.model.Address;
 import uk.gov.hmcts.reform.civil.model.CaseData;
+import uk.gov.hmcts.reform.civil.model.SolicitorServiceAddress;
 import uk.gov.hmcts.reform.civil.model.docmosis.sealedclaim.Representative;
 import uk.gov.hmcts.reform.civil.service.OrganisationService;
 import uk.gov.hmcts.reform.civil.service.flowstate.StateFlowEngine;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/RepresentativeService.java
```diff
@@ -38,11 +44,23 @@ public Representative getApplicantRepresentative(CaseData caseData) {
...
+            .serviceAddress(fromProvidedAddress(representative.getServiceAddress(), providedSolicitor1ServiceAddress))
             .build();
     }
 
+    private Address fromProvidedAddress(
+        Address registeredServiceAddress, SolicitorServiceAddress providedServiceAddress) {
+        if (providedServiceAddress != null && providedServiceAddress.getHasServiceAddress().equals(YesOrNo.YES)) {
+            return providedServiceAddress.getAddress();
+        }
+        return registeredServiceAddress;
```
pd-dudkiewicz, line 61: We could make it a bit more readable:
```
return Optional.ofNullable(providedServiceAddress)
    .map(SolicitorServiceAddress::getAddress)
    .orElse(registeredServiceAddress);
```
`hasServiceAddress` check is redundant if it's used as show condition for address field

dharmendrak: providedServiceAddress null check is just backward compatibility for releasing service code first, otherwise it will be mandatory field

adamjhc (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/RepresentativeService.java
```diff
@@ -38,11 +44,23 @@ public Representative getApplicantRepresentative(CaseData caseData) {
...
             .build();
     }
 
+    private Address fromProvidedAddress(
+        Address registeredServiceAddress, SolicitorServiceAddress providedServiceAddress) {
+        if (providedServiceAddress != null && providedServiceAddress.getRequired().equals(YesOrNo.YES)) {
+            return providedServiceAddress.getAddress();
+        }
+        return registeredServiceAddress;
+    }
```
swalker125, line 62: I would make a method for this on caseData, so that we can do caseData.getServiceAddress or something similar. Logic is then in one place rather than needed in multiple

dharmendrak: it's a generic method for three different Service addresses captured in claim creation, null check is temporary to make it backward compatible as well as it is going to be mandatory, however it can move in some common helper

swalker125 (approved):

PR #139, opened 2021-06-30
Feature/spec changes (#138)

ThirumuruganD:
Rebase all spec changes so far with master

<DO NOT MERGE - This is a Draft PR>


swalker125:
Hey guys, please follow conventional commits: https://www.conventionalcommits.org/en/v1.0.0/ for PR titles with the JIRA ticket at the end. 

example:

feat: did some stuff (CMC-100)

ThirumuruganD:
Hi @swalker125  - This is a draft PR and would require changes as needed based on PR build status.
Will update all required details before moving this to Review.

swalker125:
It might also be worth noting I do not think if statements within existing code is the way to go for this. That's going to become a nightmare to maintain very very quickly. I think we should be extracting common code but implementing callbacks separately, what do you guys think? We need to think of a way to extend the state machine too, duplication is not the way to go in my mind. Interested to know the spec teams thoughts. Sorry to be critical, it's just really important we get this right

PR #212, opened 2021-08-02
Feat cmc 1342 notify multiparty claim

HarryH96:
https://tools.hmcts.net/jira/browse/CMC-1342

### Change description ###
Deploy preview environment with Multiparty features

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/CreateClaimRespondentNotificationHandler.java
```diff
@@ -29,10 +32,15 @@
 
     private static final List<CaseEvent> EVENTS = List.of(
         NOTIFY_RESPONDENT_SOLICITOR1_FOR_CLAIM_ISSUE,
-        NOTIFY_RESPONDENT_SOLICITOR1_FOR_CLAIM_ISSUE_CC);
+        NOTIFY_RESPONDENT_SOLICITOR1_FOR_CLAIM_ISSUE_CC,
+        NOTIFY_RESPONDENT_SOLICITOR2_FOR_CLAIM_ISSUE
+    );
 
     public static final String TASK_ID = "NotifyDefendantSolicitor1";
     public static final String TASK_ID_CC = "NotifyApplicantSolicitor1CC";
```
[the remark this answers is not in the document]
HarryH96: These fields are currently being used by the production callback to determine if the notification is for the applicant solicitor (CC) or the defendant 1 solicitor. Maybe this is a backwards compatibility issue and will need to be cleaned up in a 2nd PR after the initial changes are deployed - alongside new camunda + CCD changes

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/NotifyClaimCallbackHandler.java
```diff
@@ -100,4 +155,29 @@ private SubmittedCallbackResponse buildConfirmation(CallbackParams callbackParam
...
+            .confirmationHeader(String.format(
+                "# Notification of claim sent%n## Claim number: %s",
+                caseData.getLegacyCaseReference()
+            ))
+            .confirmationBody(body)
+            .build();
+    }
+
+    protected boolean notifyBothRespondentSolicitors(CaseData caseData) {
+        return Objects.equals("Both", caseData.getDefendantSolicitorNotifyClaimOptions().getValue().getLabel());
```
ckalista, line 181: This also needs to be made backward compatible.

I suspect that as it stands, this won't work if you merge the service code first and only the service code (that is without the accompanying CCD Def changes)

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/NotifyClaimCallbackHandler.java
```diff
@@ -100,4 +155,29 @@ private SubmittedCallbackResponse buildConfirmation(CallbackParams callbackParam
...
             .build();
     }
+
+    private SubmittedCallbackResponse buildConfirmationWithSolicitorOptions(CallbackParams callbackParams) {
+        CaseData caseData = callbackParams.getCaseData();
+        String formattedDeadline = formatLocalDateTime(caseData.getClaimDetailsNotificationDeadline(), DATE_TIME_AT);
+        String body;
+
+        if (notifyBothRespondentSolicitors(caseData)) {
+            body = format(CONFIRMATION_SUMMARY, formattedDeadline) + exitSurveyContentService.applicantSurvey();
```
ckalista, line 165: Is the logic correct here in the context of backward compatibility?

We need the confirmation screen to remain unchanged until we release multi-party later down the road. This means we need the {body} to always be {CONFIRMATION_SUMMARY} until multi-party feature flag is turned on.

ckalista (changes requested):
Hiya, good job! Just a couple of questions around backward compatibility worth having a look 👍

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/CreateClaimRespondentNotificationHandler.java
```diff
@@ -62,24 +81,81 @@ private CallbackResponse notifyRespondentSolicitorForClaimIssue(CallbackParams c
...
+    //            - Divergent)
+    private CallbackResponse notifyARespondentSolicitorForClaimIssue(CallbackParams callbackParams) {
+        CaseData caseData = callbackParams.getCaseData();
+        String recipient;
+
+        CaseEvent caseEvent = CaseEvent.valueOf(callbackParams.getRequest().getEventId());
+        switch (caseEvent) {
+            case NOTIFY_RESPONDENT_SOLICITOR1_FOR_CLAIM_ISSUE:
+                recipient = caseData.getRespondentSolicitor1EmailAddress();
+                if (shouldEmailRespondent2Solicitor(caseData)) {
```
sabahirfan, line 114: This does not look right as we have a separate case for this

HarryH96: In the scenario where the user opts to Notify only the second solicitor in the Notfy Claim Options List.
This catches the users selection and sets the recipient to "Solicitor B" after this is processed, there is logic in the ```State Flow Engine``` and Camunda Diagram is configured to hand the case offline. As only 1 Solicitor is notified - the case will stay online as long as both are selected.

In camunda, this will trigger the second event to notify the other respondent solicitor :)

src/main/java/uk/gov/hmcts/reform/civil/launchdarkly/FeatureToggleService.java
```diff
@@ -37,7 +37,8 @@ public boolean isOrganisationOnboarded(String orgId) {
     }
 
     public boolean isMultipartyEnabled() {
-        return internalClient.boolVariation("multiparty", createLDUser().build(), false);
+        return true;
+        //return internalClient.boolVariation("multiparty", createLDUser().build(), false);
```
HarryH96, line 41: for testing purposes only, to be reverted

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/NotifyClaimCallbackHandler.java
HarryH96: Added the following logic for using existing `buildConfirmation` method until multiparty toggle is enabled:
```        
if (!featureToggleService.isMultipartyEnabled()
            || caseData.getDefendantSolicitorNotifyClaimOptions() == null) {
            return buildConfirmation(callbackParams);
        }
```

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/NotifyClaimCallbackHandler.java
HarryH96: This code will be evaluated by the mid event, so I have included a multiparty as part of that logic also:
```
        if (featureToggleService.isMultipartyEnabled()
            && !notifyBothRespondentSolicitors(caseData)) {
            warnings.add(WARNING_ONLY_NOTIFY_ONE_DEFENDANT_SOLICITOR);
        }
```
So warnings will only be evaluated when Multiparty is on

dharmendrak (approved):

PR #398, opened 2021-11-11
CMC-1755 MultiParty Operational RPA Workaround

jeejavr:

https://tools.hmcts.net/jira/browse/CMC-1755

### Change description ###

- Setup the new email address for multiparty
- Update to RoboticsEmailConfiguration to add new recipient for multiparty
- Updates to RoboticsNotificationService to configure and send the email
- Added new method to get the latest event trigger
- Added second claimant and second defendant in RoboticsDataMapper
- application yaml file update to add the new recipient 
- Added new test cases and updated accordingly

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/civil-service/values.yaml
+++ b/charts/civil-service/values.yaml
@@ -70,5 +70,5 @@ java:
         - robotics-notification-sender
         - robotics-notification-recipient
         - launch-darkly-sdk-key
-        - multiparty-notification-sender
-        - multiparty-notification-recipient
+        - robotics-notification-multipartyrecipient
+
--- a/charts/civil-service/values.yaml
+++ b/charts/civil-service/values.yaml
@@ -70,3 +70,5 @@ java:
         - robotics-notification-sender
         - robotics-notification-recipient
         - launch-darkly-sdk-key
+        - multiparty-notification-sender
+        - multiparty-notification-recipient
--- a/src/main/java/uk/gov/hmcts/reform/civil/service/robotics/mapper/RoboticsDataMapper.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/service/robotics/mapper/RoboticsDataMapper.java
@@ -42,11 +42,6 @@ public class RoboticsDataMapper {
     public static final String APPLICANT_ID = "001";
     public static final String RESPONDENT_ID = "002";
 
-    public static final String APPLICANT2_SOLICITOR_ID = "003";
-    public static final String RESPONDENT2_SOLICITOR_ID = "004";
-    public static final String APPLICANT2_ID = "003";
-    public static final String RESPONDENT2_ID = "004";
-
     private final RoboticsAddressMapper addressMapper;
     private final EventHistoryMapper eventHistoryMapper;
     private final OrganisationService organisationService;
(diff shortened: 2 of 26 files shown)


src/main/java/uk/gov/hmcts/reform/civil/service/robotics/mapper/RoboticsDataMapper.java
```diff
@@ -42,6 +42,11 @@
     public static final String APPLICANT_ID = "001";
     public static final String RESPONDENT_ID = "002";
 
+    public static final String APPLICANT2_SOLICITOR_ID = "003";
+    public static final String RESPONDENT2_SOLICITOR_ID = "004";
+    public static final String APPLICANT2_ID = "003";
+    public static final String RESPONDENT2_ID = "004";
```
HarryH96, line 48: @jeejavr not sure if this is relevant, do these values need to be unique?

HarryH96:
@jeejavr could you please add the ticket link + changes to the description please!

src/main/java/uk/gov/hmcts/reform/civil/service/robotics/mapper/RoboticsDataMapper.java
jeejavr: @HarryH96 not sure about that as there was a previous implementation for applicant 1 and solicitor 1. I was just updating it for applicant 2 and solicitor 2. Please let me know your thoughts on alternate implementations.

akriti-hmcts: @jeejavr  **buildSolicitors** method needs updating otherwise in MP scenario it will never build the whole solicitors list. However this change is now out of scope of this ticket as we are not attaching the json anymore to the email but would be required in one of the RPA stories we are going to look at.

akriti-hmcts: @HarryH96 These Ids are only used in the json to represent a unique object. So for example a list of solicitors will be populated within the json and each solicitor object will have its own unique Id.  See this RPA json to example to see how these ids are used - https://jsonformatter.org/60893c.

src/main/resources/application.yaml
```diff
@@ -204,6 +204,7 @@ robotics:
   notification:
     sender: hmcts.civil@gmail.com
     recipient: hmcts.civil+robotics@gmail.com
+    multipartyrecipient: damagesmultiparty@justice.gov.uk
```
akriti-hmcts, line 207: this needs changing to a test user

charts/civil-service/values.yaml
```diff
@@ -70,3 +70,5 @@ java:
         - robotics-notification-sender
         - robotics-notification-recipient
         - launch-darkly-sdk-key
+        - robotics-notification-multipartyrecipient
```
akriti-hmcts, line 73: this needs to get added to values.aat.template.yaml and values.preview.template.yaml files as well for respective environments

akriti-hmcts (changes requested):
please update the yml files as mentioned

charts/civil-service/values.yaml
jeejavr: mentioned yml files are updated

kingsley-amadu (approved):

akriti-hmcts (approved):

PR #624, opened 2022-02-01
Move idam calls to cache - its blocking master with 429 failures in logs 

mounikahmcts:
https://tools.hmcts.net/jira/browse/CMC-1873

**Root cause:** From E2E we check for ongoing business process for every event with max retries of 300. For every retry you  make a call to idam to get access token causing idam flooded with civil service calls. So, caching idam access token for 60min per user.

Screenshot 2022-02-01 at 11 56 24

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/service/UserService.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/service/UserService.java
@@ -20,4 +20,9 @@ public class UserService {
     public UserInfo getUserInfo(String bearerToken) {
         return idamClient.getUserInfo(bearerToken);
     }
+
+    @Cacheable(value = "accessTokenCache")
+    public String getAccessToken(String username, String password) {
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/hmcts/reform/civil/service/UserService.java
```diff
@@ -20,4 +20,9 @@ public UserService(IdamClient idamClient) {
     public UserInfo getUserInfo(String bearerToken) {
         return idamClient.getUserInfo(bearerToken);
     }
+
+    @Cacheable(value = "accessTokenCache")
```
mounikahmcts, line 24: Caching accessToken for 60 sec

Gareth40343 (approved):

sampankumar (approved):

akriti-hmcts (approved):

PR #662, opened 2022-02-14
Feat/civ 217 email served to claimant & defendant lr

jeejavr:
https://tools.hmcts.net/jira/browse/CIV-217

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
@@ -79,7 +79,7 @@ public class CaseData implements MappableObject {
     private final String paymentReference;
     private final DynamicList applicantSolicitor1PbaAccounts;
     private final ClaimType claimType;
-    private final SuperClaimType superClaimType;
+    private SuperClaimType superClaimType;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -79,7 +79,7 @@
     private final String paymentReference;
     private final DynamicList applicantSolicitor1PbaAccounts;
     private final ClaimType claimType;
-    private final SuperClaimType superClaimType;
+    private SuperClaimType superClaimType;
```
dtortolaV1, line 82: I'd need confirmation for this, but superClaimType is what says a claim is specified or not. I don't think this is something that can change afterwards

jeejavr: superClaimType was set specifically for UNSPEC in CaseData build() method and due to that was not able to create test cases for SPEC(please check DefendantResponseApplicantNotificationHandlerTest.java) because of that. So the only solution I could find was making superClaimType not final and setting it as UNSPEC atStateClaimDraft().

PR #672, opened 2022-02-15
Feat/civ 217 email notifications

jeejavr:
https://tools.hmcts.net/jira/browse/CIV-217

- Added new templates for claimant and defendant emails
- Added code in defendant response applicant notification handler so that correct email template, email id, and legal org name are picked up for generating email for spec journey.
- Added test cases

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/DefendantResponseApplicantNotificationHandler.java
```diff
@@ -53,15 +58,24 @@ public String camundaActivityId(CallbackParams callbackParams) {
     }
 
     private CallbackResponse notifyApplicantSolicitorForDefendantResponse(CallbackParams callbackParams) {
+
         CaseData caseData = callbackParams.getCaseData();
+        String emailTemplate = notificationsProperties.getClaimantSolicitorDefendantResponseFullDefence();
+        Map<String, String> addProperties = addProperties(caseData);
```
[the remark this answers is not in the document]
dtortolaV1: Also, flipping the superClaimType condition gets rid of the &&
The same if-else blocks suggested by Sampan can be used to initialise emailTemplate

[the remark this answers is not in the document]
jeejavr: I have added a new method addPropertiesSpec because we need to include CLAIM_LEGAL_ORG_NAME_SPEC which is only needed for spec.

src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -79,7 +79,7 @@
     private final String paymentReference;
     private final DynamicList applicantSolicitor1PbaAccounts;
     private final ClaimType claimType;
-    private final SuperClaimType superClaimType;
+    private SuperClaimType superClaimType;
```
dtortolaV1, line 82: I can't find where superClaimType is set. Why does it has to stop being final?

src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -79,7 +79,7 @@
     private final String paymentReference;
     private final DynamicList applicantSolicitor1PbaAccounts;
     private final ClaimType claimType;
-    private final SuperClaimType superClaimType;
```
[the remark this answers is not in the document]
jeejavr: previously SuperClaimType was set as UNSPEC in the CaseDataBuilder's build() method. I wasn't able to write test cases for SPEC because of this. (Please check DefendantResponseApplicantNotificationHandlerTest.java line 102)
So making SuperClaimType not final and setting SuperClaimType as UNSPEC in atStateClaimDraft()and later setting it as a SPEC claim in the test case solved the issue.

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/DefendantResponseApplicantNotificationHandler.java
```diff
@@ -75,8 +89,27 @@ private CallbackResponse notifyApplicantSolicitorForDefendantResponse(CallbackPa
...
+        );
+    }
+
     private boolean isCcNotification(CallbackParams callbackParams) {
         return callbackParams.getRequest().getEventId()
             .equals(NOTIFY_APPLICANT_SOLICITOR1_FOR_DEFENDANT_RESPONSE_CC.name());
     }
+
+    //finding legal org name
+    private String getLegalOrganisationName(CaseData caseData, CallbackParams callbackParams) {
```
mounikahmcts, line 106: If it is Applicant's org name. It worth renaming it as getApplicantsLegalOrgName

src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
dtortolaV1: you can use CaseData.toBuilder().superClaimType(newValue).build() to get a copy of the caseData with the value you want

jeejavr: that will be overwritten by the .build() at the end.

jeejavr: thank you for the suggestion, I have updated the code.

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/DefendantResponseApplicantNotificationHandler.java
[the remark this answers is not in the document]
dtortolaV1: It's not about the new method, but about the conditional. In your code you're doing
varA = methodA1(...)
varB = methodB1(...)
if (condition) {
  varA = methodA2(...)
  varB = methodB2(...)
}

methodA1 and methodB1 are called even if don't need them. Instead:

varA;
varB;
if (condition) {
  varA = methodA2(...)
  varB = methodB2(...)
} else {
  varA = methodA1(...)
  varB = methodB1(...)
}

This also makes very obvious that varA and varB initialisation is done depending on condition

PR #792, opened 2022-03-22
feat: retain solicitor references otherwise lost upon defendant response

bhssamray:
https://tools.hmcts.net/jira/browse/CMC-1985

### Change description ###
Retain solicitor references otherwise lost in defendant response journey (1v2 diff solicitor). solicitorReferences (which contains applicant solicitor reference and defendant 1 solicitor reference) was being set to null when defendant 2 responded. Similarly respondentSolicitor2Reference field was being set to null when defendant 1 was responding.
The change  done is to pick all the solicitor references from the caseDetailsBefore field in callbackParams in the callback handler method for about-to-submit of Defendant Response journey in civil-service.

All the original solicitor references values from the caseDetailsBefore are again set to the about-to-submit casedata so that no reference is lost.

```
[ ] Yes
[x ] No
```


sankaviv1 (commented):
approved

Gareth40343:
approved

douglasrice (approved):

sankaviv1 (approved):

PR #883, opened 2022-04-20
To allow a case note to be added to CCD when a claim has dropped offline

Gareth40343:
https://tools.hmcts.net/jira/browse/CMC-1886

Allowed 'Add case note' user event to be available for cases that have been taken offline.

```
[ ] Yes
[X] No
```


sankaviv1 (commented):
approved

bhssamray (commented):
Approved

PR #888, opened 2022-04-21
fix/CMC-2022 update predicate for time extension

sankaviv1:

https://tools.hmcts.net/jira/browse/CMC-2022

### Change description ###

- update timeExtension predicate to not check for response date
- fix tests

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowPredicate.java
```diff
@@ -138,8 +138,15 @@ private static boolean getPredicateForNotificationAcknowledged(CaseData caseData
     }
 
     public static final Predicate<CaseData> respondentTimeExtension = caseData ->
-        (caseData.getRespondent1TimeExtensionDate() != null && caseData.getRespondent1ResponseDate() == null)
-            || (caseData.getRespondent2TimeExtensionDate() != null && caseData.getRespondent2ResponseDate() == null);
+        getPredicateForTimeExtension(caseData);
+
+    private static boolean getPredicateForTimeExtension(CaseData caseData) {
+        if (getMultiPartyScenario(caseData) == ONE_V_TWO_TWO_LEGAL_REP) {
+            return caseData.getRespondent1TimeExtensionDate() != null
```
bhssamray, line 145: Not sure if this change of logic will affect other transitions. The condition for response date must have been for a reason

sankaviv1: I'm not sure as well. But if there's a response date and a time extension, that predicate would never be true (which seems wrong and should be split into two predicates if needed I think).

sherlynkhaw: @bhssamray talked to @sankaviv1 and think this is the right approach. 
the reason why the responseDate is removed is because if it's there it'll not appear in the event history after defendant response and we don't want that. this is mimicking the behaviour of notification acknowledge and so i think this is what it's meant to be.

sherlynkhaw (approved):

bhssamray (approved):

PR #889, opened 2022-04-21
CIV-1876 - Sending correct site id and service name to PBA.

sampankumar:
https://tools.hmcts.net/jira/browse/CIV-1876

### Change description ###
Sending correct site id and service name to payments services.

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/PaymentsService.java
```diff
@@ -36,18 +37,35 @@ private CreditAccountPaymentRequest buildRequest(CaseData caseData) {
...
+        } else if (SPEC_CLAIM.equals(caseData.getSuperClaimType())) {
+            creditAccountPaymentRequest = CreditAccountPaymentRequest.builder()
+                .accountNumber(caseData.getApplicantSolicitor1PbaAccounts().getValue().getLabel())
+                .amount(claimFee.getCalculatedAmount())
+                .caseReference(caseData.getLegacyCaseReference())
+                .ccdCaseNumber(caseData.getCcdCaseReference().toString())
+                .customerReference(customerReference)
+                .description("Claim issue payment")
+                .organisationName(organisationName)
+                .service(paymentsConfiguration.getSpecService())
```
dtortolaV1, line 64: since service and siteId are the only difference between spec and unspec, I think it'd be easier to see that if the if blocks contained only that, something like
builder = CreditAccountPaymentRequest.builder().<...commonProperties...>();
if (SPEC_CLAIM) {
  builder.service(specService).siteId(specSiteId);
} else {
  builder.service(service).siteId(siteId);
}
return builder.build();

dtortolaV1 (dismissed):

PR #931, opened 2022-04-27
Fix same unregistered solicitor scenario

Gareth40343:
https://tools.hmcts.net/jira/browse/SNI-680

For 1v2 same solicitor scenarios we always copy over respondent1OrganisationPolicy details as respondent2OrganisationPolicy details. Since in this scenario the respondent1 rep is unregistered we don't have an org policy for them so will get a null pointer when trying to copy over details. Fix is to ensure respondent1 rep is registered before trying to copy over the org policy details.

```
[ ] Yes
[X] No
```


bhssamray (commented):
approved

sankaviv1 (commented):
approved

PR #1108, opened 2022-06-15
Refactored the AssignCaseToUserHandler to use a submited event to sto…

Gareth40343:
https://tools.hmcts.net/jira/browse/CIV-2159

We have an issue where we are assigning a role to the applicant solicitor but the supplementary data added to the case from the assign api call is getting overwritten when we return an about to submit event. Seems ccd caches case data when an event is triggered and it reapplies this cached data on an about to submit, resulting in lost data changes made through an api call.
A working solution for us was to change the assign to solicitor callback handler to use a submitted event only so we don't trigger another data change in ccd after assigning a case role.

```
[ ] Yes
[X] No
```


sherlynkhaw (dismissed):

HarryH96 (dismissed):

sherlynkhaw (commented):
lgtm

akriti-hmcts (commented):
looks good

douglasrice (approved):

PR #1118, opened 2022-06-16
Email notification sent once claim is notified displays incorrect dat…

asthamalviya:
### Change description ###

https://tools.hmcts.net/jira/browse/SNI-1032

Email notification date should be issue date plus 14 days . Before it was used as issue date plus 4 months which was set as deadline for issue closed.  
ClaimNotificationDeadline is used many places and treated as deadline so cannot change the set value it will be bigger change and should be set as claimDismissedDeadline . 
But for now changed it to notification date in the template.

```
[ ] Yes
[x ] No
```


sherlynkhaw (commented):
lgtm

douglasrice (dismissed):

douglasrice (approved):

PR #1176, opened 2022-06-28
Update all documents created as part of the claimant/defendant journe…

asthamalviya:
https://tools.hmcts.net/jira/browse/CIV-2329

### Change description ###
SOT text has been added and separated as defendant and claimant text 

```
[ ] Yes
[ X] No
```


sankaviv1 (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/dq/DirectionsQuestionnaireGenerator.java
```diff
@@ -112,7 +112,7 @@ private DocmosisTemplates getDocmosisTemplate(CaseData caseData) {
                     templateId = N181_CLAIMANT_MULTIPARTY_DIFF_SOLICITOR;
                 }
                 break;
-                //FALL-THROUGH
+            //FALL-THROUGH
```
sherlynkhaw, line 115: i think this is remnant code that doesn't need to be there anymore, can be deleted pls

asthamalviya: Done

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/dq/DirectionsQuestionnaireGenerator.java
```diff
@@ -815,4 +820,21 @@ private WelshLanguageRequirements getWelshLanguageRequirements(DQ dq) {
                 welshLanguageRequirements.getDocuments()).map(Language::getDisplayedValue).orElse(""))
             .build();
     }
+
+    private String getClaimantStatementOfTruth() {
```
sherlynkhaw, line 823: I think this can be consolidated into one method and change the word to 'claimant' or 'defendant' depending on case. This way we don't need to upkeep two sets of strings and less code duplication.

_I've not tested the below so it might have some problems. There are other ways to do this too, but I'll leave it up to you to decide._

    ```
public static String getStatementOfTruth(String role) {
        return String.format("The %s believes that the facts this claim are true.\n\n\n"
            + "I am duly authorised by the %s to sign this statement.\n\n"
            + "The %s understands that proceedings for contempt of court may be brought against anyone who makes,"
            + "or causes to be made, a false statement in a document verified by a statement of truth without an honest"
            + "belief in its truth.", IntStream.range(0, 3).mapToObj(i -> role).toArray());
    }
```

sherlynkhaw: You can also keep the string as a constant on top of the file, it would make the file look tidier.

sherlynkhaw (changes requested):

sherlynkhaw (changes requested):

sherlynkhaw (approved):

sankaviv1 (dismissed):

sherlynkhaw (approved):

sabahirfan (approved):

sherlynkhaw (dismissed):

sherlynkhaw (approved):

sankaviv1 (dismissed):

sabahirfan (approved):

sankaviv1 (approved):

Gareth40343 (approved):

PR #1288, opened 2022-07-27
Feat/civ 3521

sankaviv1:

https://tools.hmcts.net/jira/browse/CIV-3521

### Change description ###

- add new case data field
- add case access category enum
- populate the field on case creation

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/enums/CaseCategory.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.hmcts.reform.civil.enums;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public enum CaseCategory {
+    UNSPEC_CLAIM,
+    SPEC_CLAIM;
```
HarryH96, line 10: Im not sure where we use this - but similar to an unspec claim, we need to set this for Specific claims.
They currently set the super claim type, which I believe is to be migrated / made redundant when we release Access Profiles service code..

Part of this PR should be to convert any usages of superClaimType - to instead use the Case Access Category field in case data

src/main/java/uk/gov/hmcts/reform/civil/enums/CaseRole.java
```diff
@@ -6,13 +6,8 @@
...
     CREATOR,
     APPLICANTSOLICITORONE,
-    APPLICANTSOLICITORTWO,
     RESPONDENTSOLICITORONE,
-    RESPONDENTSOLICITORTWO,
-    APPLICANTSOLICITORONESPEC,
-    APPLICANTSOLICITORTWOSPEC,
-    RESPONDENTSOLICITORONESPEC,
-    RESPONDENTSOLICITORTWOSPEC;
+    RESPONDENTSOLICITORTWO;
```
HarryH96, line 10: This should be reverted for now, it will be a cleanup activity for after the release of Access Profiles

PR #1313, opened 2022-08-02
civil-2930-create-refer-to-judge-event

nagaraju-deekonda:
https://tools.hmcts.net/jira/browse/CIV-2930

### Change description ###
Legal Advisor who is unable to complete a task in their work allocation assignment box  to create refer to judge event

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/civil/model/sdo/ReferToJudgeReferences.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.reform.civil.model.sdo;
```
Gulam-Hussain, line 1: I don't think this class is needed, it is trying to encompass the two ccd fields but to my knowledge they fields themselves would be need to be fields in caseData as if you try to add this class to caseData, it will look for a ccd field with this name.

Alternatively, you can make the ccd fields inside a complexType with this name and it should then capture in caseData.

m-meulendijks-v1 (approved):

src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -491,6 +491,14 @@ public boolean hasNoOngoingBusinessProcess() {
 
     private String caseManagementOrderSelection;
     private Document orderSDODocumentDJ;
+    /**
+     * RTJ = Refer To Judge.
+     */
+    private final String eventDescriptionRTJ;
```
sampankumar, line 497: Looks like we have duplicate of these 2 in CaseDataParent.java

cj1993 (approved):

sampankumar (approved):

PR #1334, opened 2022-08-12
Feat/civ 3278 court location

sherlynkhaw:
https://tools.hmcts.net/jira/browse/CIV-3278

### Change description ###

- Made court location into a dynamic list by fetching a list of courts from ref data api
- fixed e2e and api tests to reflect changes

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/CreateClaimCallbackHandler.java
```diff
@@ -123,8 +131,23 @@ public List<CaseEvent> handledEvents() {
...
+            .build();
+
         return AboutToStartOrSubmitCallbackResponse.builder()
-            .data(CaseData.builder().claimStarted(YES).build().toMap(objectMapper)).build();
+            .data(caseData.toMap(objectMapper)).build();
+    }
+
+    private List<LocationRefData> fetchLocationData(CallbackParams callbackParams) {
+        String authToken = callbackParams.getParams().get(BEARER_TOKEN).toString();
+        // Potentially renaming this method to something more generic not DJ
```
kurt-uk, line 149: think you left a task in here, needs deleting whether youre gunna rename or not

src/main/java/uk/gov/hmcts/reform/civil/model/CourtLocation.java
```diff
@@ -4,15 +4,25 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Builder;
 import lombok.Data;
+import lombok.Setter;
+import uk.gov.hmcts.reform.civil.model.common.DynamicList;
+import uk.gov.hmcts.reform.civil.model.defaultjudgment.CaseLocation;
 
+@Setter
```
kurt-uk, line 11: shouldnt need a setter as you already have data, 
data is equivalent to getter and setter
docs:
`Equivalent to @Getter @Setter @RequiredArgsConstructor @ToString @EqualsAndHashCode.`

kurt-uk (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/model/CourtLocation.java
sherlynkhaw: cool, removed! @

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/CreateClaimCallbackHandler.java
sherlynkhaw: removed it!

sankaviv1 (approved):

alberto-hortiguela (approved):

kurt-uk (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/CreateClaimCallbackHandler.java
```diff
@@ -124,8 +135,25 @@ public List<CaseEvent> handledEvents() {
...
+                   .applicantPreferredCourtLocationList(courtLocationUtils.getLocationsFromList(locations))
+                   .build());
+        }
+
         return AboutToStartOrSubmitCallbackResponse.builder()
-            .data(CaseData.builder().claimStarted(YES).build().toMap(objectMapper)).build();
+            .data(caseDataBuilder.build().toMap(objectMapper)).build();
+    }
+
+    private List<LocationRefData> fetchLocationData(CallbackParams callbackParams) {
```
dtortolaV1, line 154: This method is used here and in ResponseCallbackHandler. Can you move it to a service? It is maybe not appropriate to use callbackParams in a service method, but it can receive just the String authToken
That way it is defined once instead of twice in this PR and twice more in mine

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/CreateClaimCallbackHandler.java
```diff
@@ -432,6 +468,18 @@ private String getBody(CaseData caseData) {
...
+        // We are validating it exists on submission if not we return an error to the user.
+        if (caseData.getCourtLocation() == null
+            || caseData.getCourtLocation().getApplicantPreferredCourtLocationList() == null
+            || caseData.getCourtLocation().getApplicantPreferredCourtLocationList().getValue() == null) {
+            errorsMessages.add("Court location code is required");
+        }
+        return errorsMessages;
+    }
+
+    private List<String> validateCaseDataOld(CaseData caseData) {
```
dtortolaV1, line 482: Not important, but if these methods were, for instance validateCourtText and validateCourtChoice I think they'd be understood without the need to look at what they do

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/RespondToClaimCallbackHandler.java
```diff
@@ -151,15 +162,31 @@ private CallbackResponse populateRespondentCopyObjects(CallbackParams callbackPa
             isRespondent1 = NO;
         }
 
+        DynamicList courtLocationList = courtLocationUtils.getLocationsFromList(fetchLocationData(callbackParams));
```
dtortolaV1, line 165: this line can be moved inside the if (V1), so that it is not loaded if not needed. At least the actual call, since the value is going to be used both for r1 and r2

dtortolaV1 (commented):
Besides the added comments, we could annotate the old callback handler methods with deprecated to mark them to be eventually removed (including a comment saying that since 3278 is merged, they won't be needed)

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/CreateClaimCallbackHandler.java
sherlynkhaw: hey, these are supposed to be deleted thats why the 'old', it's only there for backwards compatibility

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/RespondToClaimCallbackHandler.java
sherlynkhaw: good spot, i'll change that

kurt-uk (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/RespondToClaimCallbackHandler.java
sherlynkhaw: actually after second thought, i think it's ok to be in there, the V1 bits would be removed almost immediately.
it'll almost be doing extra work to remove it later.

kurt-uk (approved):

dtortolaV1 (dismissed):

Gareth40343 (approved):

PR #1336, opened 2022-08-12
CIV-3024: PIP: Removing the PIN from case data

hmcts-version1-pablo:

https://tools.hmcts.net/jira/browse/CIV-3024

### Change description ###
Refactoring of the PIN creation process to a service.
Deletion of the PinUsed attribute because it is not going to be used in the end.

```
[ ] Yes
[ X] No
```


TatsyMaslova:
I do not see any logic that would check the expiry date for the pin in the post and remove the pin.. I only see logic generating the pin which in theory should have been there before this story has been started.  also in this case should the whole instance of pin in the post be removed or just the access token?

PR #1426, opened 2022-09-13
Feat/civ 4767

dtortolaV1:
CIV-4767

### Change description ###
Change text for order without hearing for small claims

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/DeadlinesCalculator.java
```diff
@@ -60,6 +60,14 @@ public LocalDate calculateFirstWorkingDay(LocalDate date) {
         return date;
     }
 
+    public LocalDate plusWorkingDays(LocalDate date, int workingDays) {
```
kurt-uk, line 63: youve missed the tests for this class

kurt-uk (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/service/DeadlinesCalculator.java
kurt-uk: also i dont think this works as the `getNextWorkingDay` method looks like it only returns the next working day if the current day is not a working day. 
so if its a monday, line 66 will return the same monday so it doesnt look like it will progress

src/test/java/uk/gov/hmcts/reform/civil/service/DeadlinesCalculatorTest.java
```diff
@@ -323,4 +324,12 @@ private BankHolidays loadFixture() throws IOException {
         String input = ResourceReader.readString("/bank-holidays.json");
         return new ObjectMapper().readValue(input, BankHolidays.class);
     }
+
+    @Test
+    public void plusWorkingDays() {
```
kurt-uk, line 329: dont need public declaration

src/test/java/uk/gov/hmcts/reform/civil/service/DeadlinesCalculatorTest.java
```diff
@@ -323,4 +324,12 @@ private BankHolidays loadFixture() throws IOException {
...
+        LocalDate start = LocalDate.of(2022, 9, 12);
+        when(nonWorkingDaysCollection.contains(start.plusDays(7))).thenReturn(true);
+        int days = 10;
+        Assertions.assertEquals(start.plusDays(15), calculator.plusWorkingDays(start, days));
```
kurt-uk, line 333: although i much prefer junit5, this class uses junit4 so it might be best to not mix the 2 in the same class and use the same assert the rest of the tests do

kurt-uk (approved):

src/test/java/uk/gov/hmcts/reform/civil/service/DeadlinesCalculatorTest.java
kurt-uk: also the test name should help describe the test scenario and should have a few that test different scenarios

dtortolaV1:
closed until agreed on a double non-prod approach

PR #1544, opened 2022-10-06
Updated JacksonCVE-2022-42003

Gareth40343:

```
[ ] Yes
[X] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -390,8 +390,8 @@ dependencies {
   implementation group: 'org.camunda.bpm', name: 'camunda-external-task-client', version: '7.16.0'
   implementation group: 'org.camunda.bpm.extension.rest', name: 'camunda-rest-client-spring-boot-starter', version: '0.0.4'
   implementation group: 'org.camunda.bpm', name: 'camunda-engine-rest-core', version: '7.16.0'
-  implementation group: 'com.fasterxml.jackson.module', name: 'jackson-module-kotlin', version: '2.13.1'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -390,8 +390,8 @@ dependencies {
   implementation group: 'org.camunda.bpm', name: 'camunda-external-task-client', version: '7.16.0'
   implementation group: 'org.camunda.bpm.extension.rest', name: 'camunda-rest-client-spring-boot-starter', version: '0.0.4'
   implementation group: 'org.camunda.bpm', name: 'camunda-engine-rest-core', version: '7.16.0'
-  implementation group: 'com.fasterxml.jackson.module', name: 'jackson-module-kotlin', version: '2.13.1'
-  implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: '2.13.1'
+  implementation group: 'com.fasterxml.jackson.module', name: 'jackson-module-kotlin', version: '2.14.0-rc1'
+  implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: '2.14.0-rc1'
```
dtortolaV1, line 395: By using the dependencies task it seems that the jackson version is constrained to 2.12.6.1 because we use spring-boot-starter-json 2.5.14
You can also check that, when you build civil-service, the jar contains (BOOT-INF/lib) the version 2.12.6.1
What is removing the throwing of the error is supressing it below, but it seems that we'll have to upgrade spring boot to actually remove the risk

PR #1566, opened 2022-10-10
Civils to use only secure document service - DM Store

alberto-hortiguela:
### Change description ###
To set document service only for civil service

```
[ ] Yes
[x] No
```


sankhajuria (approved):
Thanks, Now it wil use the secured dm store for all environments including perftest and ithc

sampankumar (approved):

HarryH96 (approved):

PR #1583, opened 2022-10-13
Feat/civ 2216 generate hearing document

miguelMolina3691:
### Change description ###

```
[ ] Yes
[ x ] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/hearing/HearingFormGenerator.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private String getFileName(CaseData caseData, DocmosisTemplates template) {
+        return String.format(template.getDocumentTitle(), caseData.getLegacyCaseReference());
+    }
+
+    private String getHearingTimeFormatted(String hearingTime) {
+        StringBuilder hearingTimeBuilder = new StringBuilder(hearingTime);
+        hearingTimeBuilder.insert(2, ':');
+        return hearingTimeBuilder.toString();
```
filippo-possenti-version1, line 96: This works on the condition that `hearingTime` is a string exactly 4 (or 7, if we use AM/PM notation) characters long.
I normally don't trust input data. I'd recommend to add a validation on the `hearingTime` string length, to ensure it's exactly as long as we expect it to be.

Also: is the time in 24h or 12h format? Are we formatting in the same 12/24 format or we want to ensure one over the other?

Given all the above I'd like to have a unit test on this function, to try all the various allowed and disallowed inputs and make sure the output is always consistent.

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/hearing/HearingFormGenerator.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private String getDateFormatted(LocalDate date) {
+        if (isNull(date)) {
+            return null;
+        } else {
```
filippo-possenti-version1, line 102: Minor: no need for the `else` here. It's automatic.

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/hearing/HearingFormGenerator.java
```diff
@@ -0,0 +1,124 @@
...
+            return date.format(DateTimeFormatter.ofPattern("dd/MMM/yyyy"));
```
filippo-possenti-version1, line 103: We appear to have a `DateFormatHelper` class. You might want to add a constant to that class for the new date format and then use the provided `DateFormatHelper::formatLocalDate` method.

I'm not pushed on using that method, but since we have it we should make sure that date formatting is handled consistently all throughout.

Also, if I'm reading Doug Keating's comments correctly, I believe this format is not among the ones he proposed... although I'm not familiar with Docmosis so I don't know how this would fit. On this part, please make sure the formatting matches the desired format and write a comment here to document the outcome.

src/main/java/uk/gov/hmcts/reform/civil/utils/HearingUtils.java
```diff
@@ -54,4 +59,55 @@ public static BigDecimal getFastTrackFee(int claimFee) {
...
+            case MINUTES_180:
+                return "3 hours";
+            case MINUTES_240:
+                return "4 hours";
+            case DAY_1:
+                return "1 day";
+            case DAY_2:
+                return "2 days";
+            default:
+                return "not defined";
```
filippo-possenti-version1, line 97: A question on this, since we're generating a template document.

Are we dealing only with english or do we expect to deal also with Welsh?

src/main/java/uk/gov/hmcts/reform/civil/utils/HearingUtils.java
```diff
@@ -54,4 +59,55 @@ public static BigDecimal getFastTrackFee(int claimFee) {
...
+        if (nonNull(hearingFee) && hearingFee.getCalculatedAmountInPence().intValue() > 0) {
+            StringBuilder builder = new StringBuilder(hearingFee.getCalculatedAmountInPence().toString());
+            builder.delete(builder.length() - 2, builder.length());
+            if (builder.length() > 3) {
+                builder.insert(1, ',');
+            }
+            builder.insert(0, '£');
+            return builder.toString();
+        }
+        return "";
```
filippo-possenti-version1, line 111: Why not using `DecimalFormat` to do the work for you?

Something like:
```
DecimalFormat formatter = new DecimalFormat("#,###.00");
formatter.format(value);
```

Also, if I'm reading the code correctly we're truncating the decimals. Are fees expected to always be whole pound values?

src/main/java/uk/gov/hmcts/reform/civil/utils/HearingUtils.java
```diff
@@ -54,4 +59,55 @@ public static BigDecimal getFastTrackFee(int claimFee) {
...
+            case VIDEO:
+                return "videoconference";
+            case TELEPHONE:
+                return "telephone";
+            default:
+                return "not defined";
+        }
+    }
+
+    public static String getHearingDuration(CaseData caseData) {
```
filippo-possenti-version1, line 76: Lots of branching in here due to the switch statement. Please add a dedicated unit test to ensure the output matches expectations for all branches.

src/test/java/uk/gov/hmcts/reform/civil/utils/HearingUtilsTest.java
```diff
@@ -60,4 +61,22 @@ void shouldReturnFee6_whenGivenAnyClaimFee() {
...
+    void shouldReturnFormattedFee2_whenGivenAnyClaimFee() {
+        assertThat(HearingUtils.formatHearingFee(
+            Fee.builder().calculatedAmountInPence(new BigDecimal(132000)).build())).isEqualTo("£1,320");
+    }
+
+    @Test
+    void shouldReturnFormattedFee3_whenGivenAnyClaimFee() {
+        assertThat(HearingUtils.formatHearingFee(
+            Fee.builder().calculatedAmountInPence(new BigDecimal(5000)).build())).isEqualTo("£50");
+    }
```
filippo-possenti-version1, line 81: These three can be turned into one `@ParameterizedTest`. Please see if you can do it.

Also, if the hearing fee allows for pences, please add relevant test parameters to test also non-whole values.

filippo-possenti-version1 (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/hearing/HearingFormGenerator.java
miguelMolina3691: What is formatted are the values of a FixedList. They are strings with the format "0800", "0830"...The format is the same as the one received from ccd. I have added a test to check the format

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/hearing/HearingFormGenerator.java
miguelMolina3691: The format is defined in the Docmosis template, this format is so that Docmosis can read the date and format it

src/main/java/uk/gov/hmcts/reform/civil/utils/HearingUtils.java
miguelMolina3691: At the moment, it is only in English

src/main/java/uk/gov/hmcts/reform/civil/utils/HearingUtils.java
miguelMolina3691: Fee values are fixed and do not contain decimals.

src/test/java/uk/gov/hmcts/reform/civil/utils/HearingUtilsTest.java
miguelMolina3691: Done

src/test/java/uk/gov/hmcts/reform/civil/utils/HearingUtilsTest.java
```diff
@@ -63,20 +67,55 @@ void shouldReturnFee7_whenGivenAnyClaimFee() {
...
+            Fee.builder().calculatedAmountInPence(new BigDecimal(0)).build())).isNull();
+    }
+
+    @ParameterizedTest
+    @ValueSource(ints = {34600, 132000, 5000})
+    void shouldReturnFormattedFee_whenGivenAnyClaimFee(int amount) {
+        BigDecimal feeAmount = new BigDecimal(amount);
+        assertThat(HearingUtils.formatHearingFee(
+            Fee.builder().calculatedAmountInPence(feeAmount).build())).isNotEmpty();
+    }
```
filippo-possenti-version1, line 99: I feel we lost something in this change. Before, with the three separate tests, we were checking for the output correctness, now we don't anymore. That's why even though you did add a `@ParameterizedTest` as I asked, you also had to leave the previous tests.

Consider the following instead:
```
    @ParameterizedTest
    @CsvSource(value = {"34600;£346", "132000;£1,320", "5000;£50"}, delimiter=';')
    void shouldReturnFormattedFee_whenGivenAnyClaimFee(int amount, String expectedOutput) {
        BigDecimal feeAmount = new BigDecimal(amount);
        assertThat(HearingUtils.formatHearingFee(
            Fee.builder().calculatedAmountInPence(feeAmount).build())).isEqualTo(expectedOutput);
    }
```

The above should check the correctness of the output both in terms of "not empty" and in terms of "formatted fee".

This should make the `shouldReturnFormattedFee_whenGivenMidClaimFee`, `shouldReturnFormattedFee_whenGivenBigClaimFee` and `shouldReturnFormattedFee_whenGivenSmallClaimFee` tests no longer needed.

hmcts-version1-ignacio (approved):

src/test/java/uk/gov/hmcts/reform/civil/utils/HearingUtilsTest.java
```diff
@@ -63,20 +67,55 @@ void shouldReturnFee7_whenGivenAnyClaimFee() {
...
+
+    @Test
+    void shouldReturnNull_whenGivenEmptyTime() {
+        assertThat(HearingUtils.getHearingTimeFormatted("")).isNull();
+    }
+
+    @Test
+    void shouldReturnTimedFormatted_whenGivenAnyTime() {
+        assertThat(HearingUtils.getHearingTimeFormatted("0500")).isEqualTo("05:00");
+    }
```
filippo-possenti-version1, line 120: One of the points of me asking for tests on this method was to make sure we check boundary conditions. In this particular test there's two (plus one) boundary conditions we need to make sure will be interpreted correctly:
- The handling of 12:00 as it could be misinterpreted as midnight (00:00)
- the check for hours above 12:00 as we want to be sure we are in 24h format
- the handling of non-whole hours to make sure minutes are considered too (just to be thorough)

Consider this test:
```
    @ParameterizedTest
    @CsvSource(value = {"0000;00:00", "0500;05:00", "1200;12:00", "2300;23:00", "1230;12:30", "2318;23:18"}, delimiter=';')
    void shouldReturnTimedFormatted_whenGivenAnyTime(String input, String expectedOutput) {
        assertThat(HearingUtils.getHearingTimeFormatted(input)).isEqualTo(expectedOutput);
    }
```

src/test/java/uk/gov/hmcts/reform/civil/utils/HearingUtilsTest.java
```diff
@@ -63,20 +67,55 @@ void shouldReturnFee7_whenGivenAnyClaimFee() {
...
+    @ParameterizedTest
+    @EnumSource(HearingDuration.class)
+    void shouldReturnHearingDuration_whenGivenAnyHearingDuration(HearingDuration hearingDuration) {
+        assertThat(HearingUtils.formatHearingDuration(hearingDuration)).isNotEmpty();
+    }
+
+    @Test
+    void shouldReturnNull_whenNotAllowedTime() {
+        assertThat(HearingUtils.getHearingTimeFormatted("50000")).isNull();
+    }
```
filippo-possenti-version1, line 110: Also strings with less than 4 characters or with text other than numbers are not allowed here.

Consider the following test instead:
```
    @ParameterizedTest
    @ValueSource(strings = {"000", "50000", "08:00", "8:00", "12:00", "23:00"})
    void shouldReturnNull_whenNotAllowedTime(String input) {
        assertThat(HearingUtils.getHearingTimeFormatted(input)).isNull();
    }
```

This test will capture also those other possible situations (even though they shouldn't happen) and will fail the test appropriately as we don't deal with them. A change to the `getHearingTimeFormatted` function goes along with this test to ensure those additional disallowed inputs are considered.

src/main/java/uk/gov/hmcts/reform/civil/utils/HearingUtils.java
```diff
@@ -94,20 +97,24 @@ public static String getHearingDuration(CaseData caseData) {
...
+
+    public static String getHearingTimeFormatted(String hearingTime) {
+        if (!isEmpty(hearingTime) && hearingTime.length() == 4) {
+            StringBuilder hearingTimeBuilder = new StringBuilder(hearingTime);
+            hearingTimeBuilder.insert(2, ':');
+            return hearingTimeBuilder.toString();
         }
-        return "";
+        return null;
     }
```
filippo-possenti-version1, line 119: Consider the following to be more strict with input values, given the "fuzzy" processing we do:

```
    public static String getHearingTimeFormatted(String hearingTime) {
        if (isEmpty(hearingTime) || hearingTime.length() != 4 || !hearingTime.matches("[0-9]+")) {
            return null;
        }

        StringBuilder hearingTimeBuilder = new StringBuilder(hearingTime);
        hearingTimeBuilder.insert(2, ':');
        return hearingTimeBuilder.toString();
    }
```

filippo-possenti-version1 (commented):
Some additional changes to improve on the tests and a specific method.
As these are not major changes to the logic but just safeties and test improvements, I'll apply them myself so you can continue focusing on your current activity without distractions.

This review is just to track the nature of the activity that I'll be carrying out.

filippo-possenti-version1 (dismissed):
The code is sound.

Provided we manage to get a successful build at some point, I'm all good with this PR.

m-meulendijks-v1 (dismissed):

filippo-possenti-version1 (approved):

m-meulendijks-v1 (dismissed):

PR #1617, opened 2022-10-17
Fix random test failure

filippo-possenti-version1:
N/A

### Change description ###
- Set the nanosecond of the date/time instance to avoid elision of trailing zeroes to cause test errors.

```
[ ] Yes
[x] No
```


filippo-possenti-version1:
As there's no way that this breaks functional tests, that the purpose of this PR is to lay the foundations for other pipeline stabilisation tasks to progress and that the part of build affected has passed, I'm making a call and merging this to allow the rest to progress.

PR #1637, opened 2022-10-19
Fix flaky unit test and refactor

filippo-possenti-version1:
N/A

### Change description ###
- Fix: unit test random failure due to date/time conversion
- New: JacksonConfiguration to define how date/time conversion is supposed to be handled by jackson (prevents date/time values to have variable lengths depending on second and microsecond values)
- Other: minor refactor of test class to reduce amount of linting issues

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/config/JacksonConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/config/JacksonConfiguration.java
@@ -11,6 +11,7 @@ import org.springframework.context.annotation.Configuration;
 
 @Configuration
 public class JacksonConfiguration {
+
     public static final String DATE_FORMAT = "yyyy-MM-dd";
     public static final String DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss";
 
--- a/src/main/java/uk/gov/hmcts/reform/civil/config/JacksonConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/config/JacksonConfiguration.java
@@ -1,0 +1,28 @@
+package uk.gov.hmcts.reform.civil.config;
+
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+
+import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateSerializer;
+import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateTimeSerializer;
+import org.springframework.boot.autoconfigure.jackson.Jackson2ObjectMapperBuilderCustomizer;
+import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/civil/config/JacksonConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateTimeSerializer;
+import org.springframework.boot.autoconfigure.jackson.Jackson2ObjectMapperBuilderCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+
+@Configuration
+public class JacksonConfiguration {
```
filippo-possenti-version1, line 13: This is the class I want people to be aware of during review.

Its introduction defines a clear behaviour upon date serialization that will result in all date/time values to be serialized the same way.

Previously we would have had different formats based on the value of the date, as in these examples:
[2022-10-19 15.36.00.000000] -> "2022-10-19T15.36"
[2022-10-19 15.36.01.000000] -> "2022-10-19T15.36.01"
[2022-10-19 15.36.01.000001] -> "2022-10-19T15.36.01.000001"

With this change we will always have the same behaviour (for date/time, but similarly for dates, albeit with a different format):
[2022-10-19 15.36.00.000000] -> "2022-10-19T15.36.00"
[2022-10-19 15.36.01.000000] -> "2022-10-19T15.36.01"
[2022-10-19 15.36.01.000001] -> "2022-10-19T15.36.01"

The reason I'd like people to review this carefully is because although probably not breaking this change will affect the entire application.

src/test/java/uk/gov/hmcts/reform/civil/handler/callback/user/NotSuitableSDOCallbackHandlerTest.java
```diff
@@ -60,42 +61,28 @@ public class NotSuitableSDOCallbackHandlerTest extends BaseCallbackHandlerTest {
...
-                             0);  // set to avoid elision of zeroes to cause random test errors.
-
-            given(time.now()).willReturn(startedDate);
+            given(time.now()).willReturn(LocalDateTime.now());
 
         }
 
-        @SuppressWarnings("unchecked")
         @Test
         void checkUnsuitableSDODate() {
```
filippo-possenti-version1, line 81: This is the test that was causing problems. With the JacksonConfiguration class now this test can be simplified.

HarryH96 (approved):

cj1993 (approved):

PR #1673, opened 2022-10-26
Feat/civ 2736 - Trial Ready Check for WA

m-meulendijks-v1:
https://tools.hmcts.net/jira/browse/CIV-2736

### Change description ###
Trial ready check backend classes.

```
[ ] Yes
[x] No
```


hmcts-version1-ignacio (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/TrialReadyCheckHandler.java
```diff
@@ -0,0 +1,35 @@
...
+    public void handleTask(ExternalTask externalTask) {
+        List<CaseDetails> cases = caseSearchService.getCases();
+        log.info("Job '{}' found {} case(s)", externalTask.getTopicName(), cases.size());
+
+        cases.forEach(caseDetails -> {
+            try {
+                applicationEventPublisher.publishEvent(new TrialReadyCheckEvent(caseDetails.getId()));
+            } catch (Exception e) {
+                log.error("Updating case with id: '{}' failed", caseDetails.getId(), e);
+            }
```
filippo-possenti-version1, line 32: A question here:
is it intentional that we log the exception and then swallow it continuing to the next one or are we meant to re-throw it?

m-meulendijks-v1: Good question - the other scheduler handlers were just logging, so I followed that pattern.
However, some other handlers are explicitly throwing errors, so perhaps it would be good to follow that example?
And then we should probably change the old handlers as well to bring everything into line.

filippo-possenti-version1: I'm inclined to throw the error, unless there's a proven reason to swallow it. Should that reason arise in the future, we can always change the code, whereas swallowing the error makes for unstable systems that don't get diagnosed.

Please make the change.

filippo-possenti-version1 (changes requested):
Marking as "Request Changes" to cater for that exception swallowing.

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/TrialReadyCheckHandler.java
filippo-possenti-version1: Actually: leave that with me and keep focusing on your current ticket. I'll be putting up a commit to throw that exception momentarily.

filippo-possenti-version1: Okay, we'll leave it as is and I'll pick it up separately.

filippo-possenti-version1 (approved):

PR #1706, opened 2022-11-01
Disable polling for all envs

akriti-hmcts:
Background - Currently civil holds a polling mechanism to recover from the failures however currently the internal exceptions thrown by civil are retried by polling every 5 mins causing our logs to be full and making it hard to debug things. The poller fails to serve it's purpose at the moment.

As a part of this change -
1. Defaulting polling to be disabled in all environments including PROD until this ticket is addressed and we are in a position of re enabling the polling - (https://tools.hmcts.net/jira/browse/CIV-5365)
2. Updating the env variables for polling class to align with what is used in java classes.

```
[ ] Yes
[x ] No
```


HarryH96 (dismissed):

sampankumar (dismissed):

charts/civil-service/values.aat.template.yaml
```diff
@@ -61,7 +61,7 @@ java:
     FEIGN_CLIENT_CONFIG_REMOTEREPOSITORYSERVICE_URL: http://${SERVICE_NAME}-camunda/engine-rest/
     SERVICE_REQUEST_UPDATE: http://${SERVICE_FQDN}/service-request-update
     TESTING_SUPPORT_ENABLED: true
-    POLLING_EVENT_EMITTER_ENABLED: false
+    polling.event.emitter.enabled: false
```
filippo-possenti-version1, line 64: Is there a "less invasive" way to do this? Like by commenting the line explaining why it's been commented instead of being taken out entirely?

akriti-hmcts: it is not taken out @filippo-possenti-version1 - the property name was not used correctly so that is why always defaulting to true, I have changed just the name

filippo-possenti-version1: Got it. All good then.

filippo-possenti-version1 (approved):

charts/civil-service/values.aat.template.yaml
adusumillipraveen, line 64: Earlier name is a better naming convention for env vars. Spring Relaxed binding should honour ur earlier value and infact the the recommended naming convention (Upper case with _ )

akriti-hmcts: agreed Praveen, then this will require further mapping of the property, will do

sampankumar (approved):

adusumillipraveen (approved):

sabahirfan (approved):

PR #1710, opened 2022-11-01
Fix more flaky tests

filippo-possenti-version1:
N/A

### Change description ###
- fixed more tests failing due to the time of the day

```
[ ] Yes
[ ] No
```


filippo-possenti-version1:
As this is a trivial one, won't break anything at all and there's a build failing because of the problem that this would fix, I'll make a judgment call and merge this regardless of approvals.

PR #1748, opened 2022-11-08
allow migration events always

sankaviv1:
### Change description ###

- don't evaluate state flow engine for migration event

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/aspect/EventAllowedAspect.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/aspect/EventAllowedAspect.java
@@ -56,9 +56,7 @@ public class EventAllowedAspect {
             stateHistoryBuilder.append(s.getName());
             stateHistoryBuilder.append(", ");
         });
-        if (flowStateAllowedEventService.isAllowed(caseDetails, caseEvent)
-                || caseEvent.equals(CaseEvent.migrateCase)
-        ) {
+        if (flowStateAllowedEventService.isAllowed(caseDetails, caseEvent)) {
             return joinPoint.proceed();
         } else {
             log.info(format(
--- a/src/main/java/uk/gov/hmcts/reform/civil/aspect/EventAllowedAspect.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/aspect/EventAllowedAspect.java
@@ -50,18 +50,16 @@ public class EventAllowedAspect {
         CaseEvent caseEvent = CaseEvent.valueOf(callbackParams.getRequest().getEventId());
         CaseDetails caseDetails = callbackParams.getRequest().getCaseDetails();
         CaseData caseData = callbackParams.getCaseData();
-        StringBuilder stateHistoryBuilder = new StringBuilder();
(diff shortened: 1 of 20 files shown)


HarryH96 (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowStateAllowedEventService.java
```diff
@@ -963,6 +963,10 @@ public boolean isAllowedOnStateForSpec(String stateFullName, CaseEvent caseEvent
     public boolean isAllowed(CaseDetails caseDetails, CaseEvent caseEvent) {
         CaseData caseData = caseDetailsConverter.toCaseData(caseDetails);
 
+        if (caseEvent.equals(migrateCase)) {
```
sabahirfan, line 966: would be nice to have a unit test for this as well

HarryH96 (approved):

src/main/java/uk/gov/hmcts/reform/civil/aspect/EventAllowedAspect.java
```diff
@@ -51,11 +51,14 @@ public Object checkEventAllowed(
         CaseDetails caseDetails = callbackParams.getRequest().getCaseDetails();
         CaseData caseData = callbackParams.getCaseData();
         StringBuilder stateHistoryBuilder = new StringBuilder();
-        FlowState flowState = flowStateAllowedEventService.getFlowState(caseData);
-        stateFlowEngine.evaluate(caseData).getStateHistory().forEach(s -> {
-            stateHistoryBuilder.append(s.getName());
-            stateHistoryBuilder.append(", ");
-        });
+        FlowState flowState = FlowState.fromFullName("MAIN.DRAFT");
```
akriti-hmcts, line 54: This would define a draft state for all events then will overwrite for non migration events. For better clarity suggest to use -

  if (caseEvent.equals(CaseEvent.migrateCase)) {
FlowState flowState = FlowState.fromFullName("MAIN.DRAFT");
} else {
Flowstate flowState = flowStateAllowedEventService.getFlowState(caseData);
            stateFlowEngine.evaluate(caseData).getStateHistory().forEach(s -> {
                stateHistoryBuilder.append(s.getName());
                stateHistoryBuilder.append(", ");
            });
        }

akriti-hmcts (commented):
minor suggestion

src/main/java/uk/gov/hmcts/reform/civil/aspect/EventAllowedAspect.java
sankaviv1: @akriti-hmcts  i tried this but because the flowState variable is used only for the logs below, it would still have to be initialized outside of the if statements. instead i moved this code block closer to the logs so we don't have to initialize it with any value.

src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowStateAllowedEventService.java
sankaviv1: added test

HarryH96 (approved):

PR #1754, opened 2022-11-09
logging times when processing SBP callback

HarryH96:
N/A

### Change description ###
- Start Business Process handler is taking up to 7 minutes to complete based on camunda logs
- This is having knock on impacts to the service (triggering multiple processes as the Business Process Flag is not set to Started)
- Logging timings at various stages throughout the handler to try pinpoint delays in execution

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/businessprocess/StartBusinessProcessCallbackHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/businessprocess/StartBusinessProcessCallbackHandler.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.reform.civil.handler.callback.camunda.businessprocess;
 
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.ccd.client.model.AboutToStartOrSubmitCallbackResponse;
 import uk.gov.hmcts.reform.ccd.client.model.CallbackResponse;
@@ -19,6 +20,7 @@ import static uk.gov.hmcts.reform.civil.callback.CallbackType.ABOUT_TO_SUBMIT;
 import static uk.gov.hmcts.reform.civil.callback.CaseEvent.START_BUSINESS_PROCESS;
 
 @Service
+@Slf4j
 @RequiredArgsConstructor
 public class StartBusinessProcessCallbackHandler extends CallbackHandler {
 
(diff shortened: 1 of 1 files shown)


sampankumar (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/businessprocess/StartBusinessProcessCallbackHandler.java
```diff
@@ -38,14 +40,19 @@ public List<CaseEvent> handledEvents() {
     private final CaseDetailsConverter caseDetailsConverter;
 
     private CallbackResponse startBusinessProcess(CallbackParams callbackParams) {
+        log.info("Starting callback for start business process. Time started {}:", System.nanoTime());
```
sabahirfan, line 43: log statements are usually prefixed with the timestamp. you may want to just calculate the time taken for the method execution. i.e. 

```suggestion
long start = System.nanoTime();
// ...
long finish = System.nanoTime();
long timeElapsed = finish - start;
log.info("Tim taken for start business process {}:", timeElapsed);
```

HarryH96: I want to identify where specifically during execution there is a delay. We can tell from Camunda that the business process task is taking 7-15 minutes to complete as Camunda logs tell us:
- When its emitted or started the external task
- When it has finished the external task

In theory, I could remove the timestamps from all logs, and just have a statement - as you say the log will be timestamped on azure logs

HarryH96: I could in theory create a stopwatch and log various times when its processing? Open to suggestions

sabahirfan: > I want to identify where specifically during execution there is a delay. We can tell from Camunda that the business process task is taking 7-15 minutes to complete as Camunda logs tell us:
> 
> * When its emitted or started the external task
> * When it has finished the external task
> 
> In theory, I could remove the timestamps from all logs, and just have a statement - as you say the log will be timestamped on azure logs

This callback is simply adding an entry in Map and returning response back to CCD. This should be pretty fast, can't imagine it taking more than few milliseconds. 

You cam simply add log message here, if that helps debugging, and timestamp will be prefixed automatically.

PR #1848, opened 2022-11-24
Not changing the state to Judicial Referral anymore in java

sampankumar:
### Change description ###
Not changing the state to Judicial Referral anymore in java

```
[ ] Yes
[ ] No
```


Jenkinsfile_CNP
```diff
@@ -100,7 +101,7 @@ withPipeline(type, product, component) {
       ./bin/add-roles.sh
       ./bin/pull-latest-release-asset.sh civil-ccd-definition civil-ccd-definition.zip
       ./bin/pull-latest-release-asset.sh civil-camunda-bpmn-definition civil-camunda-bpmn-definition.zip
-      ./bin/import-ccd-definition.sh "-e *-prod.json,*HNL-nonprod.json"
+      ./bin/import-ccd-definition.sh "-e *-prod.json,*HNL-nonprod.json" ${ccddefbranch}
```
m-meulendijks-v1, line 104: Does this change need to be here?

sampankumar: I will leave it there so that we can use it in other cases, but I will make the default branch as master

m-meulendijks-v1 (approved):

PR #1854, opened 2022-11-25
CIV:4173 added additional fields to objects for docmosis and added new template

kurt-uk:

https://tools.hmcts.net/jira/browse/CIV-4173

### Change description ###
extended the expert, party and witness object to have the phone and email and preemted the firstname lastname fields coming in in the dq uplift stories.

added the new template behind hnl toggle

## Screenshot
Screenshot 2022-11-25 at 12 00 53

```
[ ] Yes
[x] No
```


asthamalviya (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/dq/DirectionsQuestionnaireGenerator.java
```diff
@@ -744,9 +775,14 @@ private Experts getSmallClaimExperts(DQ dq, CaseData caseData) {
         Expert expertDetails;
         if (experts != null) {
             expertDetails = Expert.builder()
-                .name(experts.getExpertName())
+                .name(String.format("%s %s", experts.getFirstName(), experts.getLastName()))
```
Gareth40343, line 778: I think this logic needs to be behind a hnl toggle and if hnl is not enabled you should still be mapping experts.getExpertName().

As it stands if this is merged in and hnl toggle is off in prod it will result in dq being populated with blank names, since in prod it will still be capturing the name under the expertName field until hnl is enabled.

sherlynkhaw: i didn't cr the whole pr but this part caught my eye.
i noticed that in your document that you're using for hnl you have firstname lastname instead of just name.
if thats the case, we can just not make any changes to name - so you don't upset the prod cases.

```.name(experts.getExpertName())```

LTurkingtonHMCTS (dismissed):

sankaviv1 (approved):

LTurkingtonHMCTS (dismissed):

sankaviv1 (approved):

Gareth40343 (approved):

HarryH96 (approved):

asthamalviya (approved):

PR #1877, opened 2022-11-29
Feat/civ 5364

bhagyashreesharma90:
https://tools.hmcts.net/jira/browse/CIV-5364

### Change description ###
Added new exception class to handle non retriable errors

```
[ ] Yes
[ ] No
```


hmcts-version1-ignacio (approved):

drummondjm (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/ClaimDismissedHandler.java
```diff
@@ -28,6 +28,7 @@ public void handleTask(ExternalTask externalTask) {
             try {
                 applicationEventPublisher.publishEvent(new DismissClaimEvent(caseDetails.getId()));
             } catch (Exception e) {
+                //log exception and continue with other cases
```
filippo-possenti-version1, line 31: As I'm the one who asked to put these comments:
what will happen in terms of logic in the code is fairly obvious. The comment should make it clear that this is on purpose.

I would suggest rephrasing the comment as something on the line of:
`//we continue handling other cases on purpose, as we don't want to stop processing of valid cases just because an error happened on one`

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/HearingFeeDueHandler.java
```diff
@@ -49,6 +49,7 @@ public void handleTask(ExternalTask externalTask) {
                     applicationEventPublisher.publishEvent(new HearingFeeUnpaidEvent(caseDetails.getId()));
                 }
             } catch (Exception e) {
+                //log exception and continue with other cases
```
filippo-possenti-version1, line 52: As I'm the one who asked to put these comments:
what will happen in terms of logic in the code is fairly obvious. The comment should make it clear that this is on purpose.

I would suggest rephrasing the comment as something on the line of:
`//we continue handling other cases on purpose, as we don't want to stop processing of valid cases just because an error happened on one. We log the error to leave a trace that something may need to be looked into`

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/StartGeneralApplicationBusinessProcessTaskHandler.java
```diff
@@ -51,31 +56,40 @@ public VariableMap getVariableMap() {
...
                                                                   ExternalTaskInput.class);
-        String caseId = externalTaskInput.getCaseId();
-        CaseEvent caseEvent = externalTaskInput.getCaseEvent();
-        StartEventResponse startEventResponse = coreCaseDataService.startUpdate(caseId, caseEvent);
-        CaseData data = caseDetailsConverter.toCaseData(startEventResponse.getCaseDetails());
-        List<Element<GeneralApplication>> generalApplications = data.getGeneralApplications();
+        } catch (ValueMapperException | IllegalArgumentException e) {
+            throw new InvalidCaseDataException("Mapper conversion failed due to incompatible types", e);
+        }
+        if (null != externalTaskInput) {
```
filippo-possenti-version1, line 66: This `if` shouldn't be necessary. As the exception gets catched and re-thrown in theory that value can never be `null`.

If the compiler or SonarLint are complaining, try using `Objects.requireNonNull`.

```
        try {
            externalTaskInput = Objects.requireNonNull(mapper.convertValue(externalTask.getAllVariables(),
                                                                  ExternalTaskInput.class));
        } catch (ValueMapperException | IllegalArgumentException | NullPointerException e) {
            throw new InvalidCaseDataException("Mapper conversion failed due to incompatible types", e);
        }
```
Even better if you put the `NullPointerException` in its own `catch` clause as it's not really a matter of incompatible types but of data not being provided altogether.

hmcts-version1-ignacio (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/TakeCaseOfflineHandler.java
```diff
@@ -29,6 +29,7 @@ public void handleTask(ExternalTask externalTask) {
                 log.info("Current case status '{}'", caseDetails.getState());
                 applicationEventPublisher.publishEvent(new TakeCaseOfflineEvent(caseDetails.getId()));
             } catch (Exception e) {
+                //log exception and continue with other cases
```
filippo-possenti-version1, line 32: As I'm the one who asked to put these comments:
what will happen in terms of logic in the code is fairly obvious. The comment should make it clear that this is on purpose.

I would suggest rephrasing the comment as something on the line of:
`//we continue handling other cases on purpose, as we don't want to stop processing of valid cases just because an error happened on one. We log the error to leave a trace that something may need to be looked into`

filippo-possenti-version1 (changes requested):
None of the things I commented is a dealbreaker, but they would all help making the code clearer. Let's make those changes.

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/HearingFeeDueHandler.java
bhagyashreesharma90: done

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/TakeCaseOfflineHandler.java
bhagyashreesharma90: done

filippo-possenti-version1 (approved):

drummondjm (dismissed):

src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/BaseExternalTaskHandler.java
```diff
@@ -43,6 +44,10 @@ default void execute(ExternalTask externalTask, ExternalTaskService externalTask
         } catch (BpmnError e) {
             externalTaskService.handleBpmnError(externalTask, e.getErrorCode());
             log.error("Bpmn error for external task '{}' with processInstanceId '{}'",
+                      topicName, processInstanceId, e
+            );
+        } catch (NotRetryableException e) {
```
akriti-hmcts, line 49: add StateFlowException to the NotRetryableException too

akriti-hmcts (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/exceptions/CaseIdNotProvidedException.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public CaseIdNotProvidedException() {
+        super(ERROR_MESSAGE);
+    }
+
+    public CaseIdNotProvidedException(String message) {
+        super(message);
+    }
+
+    public CaseIdNotProvidedException(String message, Throwable cause) {
```
akriti-hmcts, line 15: I don't think there is a need of CaseIdNotProvidedException as this will always be there, once a case is initiated this is the generated straight away in the system

filippo-possenti-version1: This is provided as SonarLint complains of some "possible" null values. This situation will probably never happen, but if it does there's definitely something wrong with the data.

We can probably remove the dedicated exception without sacrificing functionality. I'll take a look.

akriti-hmcts (approved):

filippo-possenti-version1 (approved):

PR #1908, opened 2022-12-02
turning off poller for all the enviornments

asthamalviya:
### Change description ###
Switching off the poller for all the environments as it was already off for lower environments after some discussions decided to off it for production as well

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/PollingEventEmitterHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/handler/tasks/PollingEventEmitterHandler.java
@@ -17,7 +17,7 @@ import static java.lang.String.format;
 @Slf4j
 @RequiredArgsConstructor
 @Component
-@ConditionalOnExpression("${polling.event.emitter.enabled:true}")
+@ConditionalOnExpression("${polling.event.emitter.enabled:false}")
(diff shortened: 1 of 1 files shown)


src/main/resources/application.yaml
```diff
@@ -243,6 +243,11 @@ exit-survey:
   applicant-link: https://www.smartsurvey.co.uk/s/CivilDamages_ExitSurvey_Claimant/
   respondent-link: https://www.smartsurvey.co.uk/s/CivilDamages_ExitSurvey_Defendant/
 
+polling:
+  event:
+    emitter:
+      enabled: ${POLLING_EVENT_EMITTER_ENABLED:false}
+
```
hgulv1, line 250: im confused as to what the change is here? you just moved a few lines above?

hgulv1 (changes requested):

src/main/resources/application.yaml
sabahirfan: Previously this was for only idam profile and not for default. Hence it was not being picked up in prod. So we have to move in default profile.

sabahirfan (approved):

HarryH96 (approved):

Gareth40343 (dismissed):

hgulv1 (dismissed):

sankaviv1 (approved):

PR #1890, opened 2022-12-01
Feat/civ 4822 spec state flow test

drummondjm:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/CIV-4822

Stateflow specified testing 

StateFlowEngineSpecTest.java
CaseDataBuilderSpec.java
**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ x] No
```

PR #1947, opened 2022-12-08
Revert "Feat/civ 4822 spec state flow test"

filippo-possenti-version1:
Reverts hmcts/civil-service#1890


filippo-possenti-version1:
Reverting, due to tests failing in master.

PR #2018, opened 2022-12-20
Change the logic on nextDeadline (4Months) calculation

asthamalviya:
https://tools.hmcts.net/jira/browse/CIV-6545

### Change description ###

Added logic that if newly calculated claimDetailsNotificationDeadline is crossing 4month deadline which is calculated and saved as ClaimNotificationDeadline/nextDeadline in previous step under  ClaimIssueCallbackHandler then the claimDetailsNotificationDeadline/nextDeadline should be same as ClaimNotificationDeadline 

For testing purpose I used variable claimDetailsNotificationDeadline in the logic rather than nextDeadline. But both the values were set same when claim was issued.

```
[ ] Yes
[x] No
```


sankaviv1 (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/NotifyClaimCallbackHandler.java
```diff
@@ -190,6 +190,11 @@ private CallbackResponse submitClaim(CallbackParams callbackParams) {
             claimDetailsNotificationDeadline = getDeadline(claimNotificationDate);
         }
 
+        if (claimDetailsNotificationDeadline.isAfter(caseData.getClaimNotificationDeadline())
```
sampankumar, line 193: I can see that the code just above this is setting the claimDetailsNotificationDeadline based on feature flag.

` if (featureToggleService.isCertificateOfServiceEnabled() && areAnyRespondentsLitigantInPerson(caseData)) {
            claimDetailsNotificationDeadline = getDeadline(getServiceDate(caseData));
        } else {
            claimDetailsNotificationDeadline = getDeadline(claimNotificationDate);
        }`

Your code is going to overwrite the deadline date no matter what feature flag's value is. Is that what is required?

asthamalviya: This logic is for calculated date weather its calculated by COS or else block . In case that date is crossing 4 months deadline then it should be overwrite and not cross 4 month deadline

sampankumar (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/NotifyClaimCallbackHandler.java
```diff
@@ -190,6 +190,11 @@ private CallbackResponse submitClaim(CallbackParams callbackParams) {
...
+            || claimDetailsNotificationDeadline.isEqual(caseData.getClaimNotificationDeadline())) {
+            claimDetailsNotificationDeadline = caseData.getClaimNotificationDeadline();
+        }
```
akriti-hmcts, line 196: why do you need an if block here, why it can't be simply -
claimDetailsNotificationDeadline = caseData.getClaimNotificationDeadline();
All we need is claimDetailsNotificationDeadline to be same as claimNotificationDeadline in any case right?

asthamalviya: No Akriti we need to check if we are crossing the 4 month then only we want claimDetailsNotificationDeadline = claimNotificationDeadline  otherwise we will add 14 days and that can be different if claimDetailsNotificationDeadline is much less than 4 month end point

HarryH96 (approved):

PR #2271, opened 2023-02-10
CIV-6994 removed PREPARE_FOR_HEARING_CONDUCT_HEARING event, case need triggers…

drummondjm:
https://tools.hmcts.net/jira/browse/CIV-6994

allows relisting of hearing notice by changing state back to case progression

```
[ ] Yes
[ x] No
```


maciejmodzelewski (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/HearingScheduledHandler.java
```diff
@@ -205,17 +205,15 @@ private CallbackResponse getDueDateAndFee(CallbackParams callbackParams) {
...
+                .state(HEARING_READINESS.name())
                 .data(caseDataBuilder.build().toMap(objectMapper))
                 .build();
         } else {
-            state = "PREPARE_FOR_HEARING_CONDUCT_HEARING";
-            caseDataBuilder.businessProcess(BusinessProcess.ready(PREPARE_FOR_HEARING_CONDUCT_HEARING));
+            caseDataBuilder.businessProcess(BusinessProcess.ready(HEARING_SCHEDULED));
             return AboutToStartOrSubmitCallbackResponse.builder()
-                .state(state)
+                .state(PREPARE_FOR_HEARING_CONDUCT_HEARING.name())
```
akriti-hmcts, line 216: is this a business requirement that the case will be set to a different state (and not HEARING_READINESS) if it is a relisting?

filippo-possenti-version1: The change made here was merely to amend the `caseDataBuilder.businessProcess` call. This change you highlighted was just a minor code refactoring and could be reverted in case.

I do however see your point: the decision of whether using PREPARE_FOR_HEARING_CONDUCT_HEARING or HEARING_READINESS should be subject to whether the fee has been paid or not and whether it's been paid fully, as otherwise we may involuntarily waive the fee or get the case stuck.

That said, the requirements state:
"As a listing officer

If I am re-listing a small claims hearing or fast track trial I need to be able to remove/delete any reference to hearing fee as this will have already been paid"

The above effectively implies that (at least for the purposes of this tactical solution) it is assumed that in case of a re-listing the fee has already been paid, which therefore means it's correct that the state is put in PREPARE_FOR_HEARING_CONDUCT_HEARING.

Still, I'll verify this separately with Doug to make sure that this was intended and is not the result of the intrinsic ambiguity of natural language.

filippo-possenti-version1 (approved):

PR #2403, opened 2023-03-07
CIV-6352 fix shortcircuiting issue

bhagyashreesharma90:
https://tools.hmcts.net/jira/browse/CIV-6352

### Change description ###
EvidenceUploadApplicantHandler.java - Fix for Short circuiting
EvidenceUploadHandlerBase.java -  Fix for Short circuiting
EvidenceUploadRespondentHandler.java - Fix for Short circuiting
BundleCreationTriggerEventHandler.java -Fix for Short circuiting and added one filter to filter bundles with null bundlehearingdate (could be possible as same case data object is used by other existing civil stitching service)
BundleRequestMapper.java - Fix for Short circuiting and one if condition
EvidenceUploadApplicantHandlerTest.java - Fix tests
EvidenceUploadRespondentHandlerTest.java - Fix tests
 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/EvidenceUploadHandlerBase.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/EvidenceUploadHandlerBase.java
@@ -355,57 +355,60 @@ abstract class EvidenceUploadHandlerBase extends CallbackHandler {
     void addUploadDocList(List<Element<UploadEvidenceDocumentType>> uploadedEvidenceAfterBundle,
                                   List<Element<UploadEvidenceDocumentType>> documentUploadEvidenceType,
                                   Optional<Bundle> bundleDetails, String docType) {
-        if (null != documentUploadEvidenceType) {
-            documentUploadEvidenceType.forEach(uploadEvidenceDocumentType -> {
-                if (null != uploadEvidenceDocumentType.getValue().getCreatedDatetime()
-                    && bundleDetails.get().getCreatedOn().isPresent()
-                    && uploadEvidenceDocumentType.getValue().getCreatedDatetime()
-                    .isAfter(bundleDetails.get().getCreatedOn().get())) {
-                    uploadedEvidenceAfterBundle.add(ElementUtils.element(UploadEvidenceDocumentType.builder()
-                                                                             .typeOfDocument(docType)
-                                                                             .createdDatetime(uploadEvidenceDocumentType.getValue().getCreatedDatetime())
-                                                                             .documentUpload(uploadEvidenceDocumentType.getValue().getDocumentUpload())
-                                                                             .build()));
-                }
-            });
+        if (null == documentUploadEvidenceType) {
+            return;
         }
+        documentUploadEvidenceType.forEach(uploadEvidenceDocumentType -> {
+            if (null != uploadEvidenceDocumentType.getValue().getCreatedDatetime()
+                && bundleDetails.get().getCreatedOn().isPresent()
+                && uploadEvidenceDocumentType.getValue().getCreatedDatetime()
+                .isAfter(bundleDetails.get().getCreatedOn().get())) {
+                uploadedEvidenceAfterBundle.add(ElementUtils.element(UploadEvidenceDocumentType.builder()
+                                                                         .typeOfDocument(docType)
+                                                                         .createdDatetime(uploadEvidenceDocumentType.getValue().getCreatedDatetime())
+                                                                         .documentUpload(uploadEvidenceDocumentType.getValue().getDocumentUpload())
+                                                                         .build()));
+            }
+        });
     }
 
     void addWitnessDocList(List<Element<UploadEvidenceDocumentType>> uploadedEvidenceAfterBundle,
                                    List<Element<UploadEvidenceWitness>> documentUploadEvidenceType,
                                    Optional<Bundle> bundleDetails, String docType) {
-        if (null != documentUploadEvidenceType) {
-            documentUploadEvidenceType.forEach(uploadEvidenceDocumentTypeElement -> {
-                if (uploadEvidenceDocumentTypeElement.getValue().getCreatedDatetime().isAfter(bundleDetails.get().getCreatedOn().orElse(
-                    null))) {
-                    uploadedEvidenceAfterBundle.add(ElementUtils.element(UploadEvidenceDocumentType.builder()
-                                                                             .typeOfDocument(docType)
-                                                                             .createdDatetime(
-                                                                                 uploadEvidenceDocumentTypeElement.getValue().getCreatedDatetime())
-                                                                             .documentUpload(
-                                                                                 uploadEvidenceDocumentTypeElement.getValue().getWitnessOptionDocument())
-                                                                             .build()));
-                }
-            });
+        if (null == documentUploadEvidenceType) {
+            return;
         }
+        documentUploadEvidenceType.forEach(uploadEvidenceDocumentTypeElement -> {
+            if (uploadEvidenceDocumentTypeElement.getValue().getCreatedDatetime().isAfter(bundleDetails.get().getCreatedOn().orElse(
+                null))) {
+                uploadedEvidenceAfterBundle.add(ElementUtils.element(UploadEvidenceDocumentType.builder()
+                                                                         .typeOfDocument(docType)
+                                                                         .createdDatetime(
+                                                                             uploadEvidenceDocumentTypeElement.getValue().getCreatedDatetime())
+                                                                         .documentUpload(
+                                                                             uploadEvidenceDocumentTypeElement.getValue().getWitnessOptionDocument())
+                                                                         .build()));
+            }
+        });
     }
 
     void addExpertDocList(List<Element<UploadEvidenceDocumentType>> uploadedEvidenceAfterBundle,
                                   List<Element<UploadEvidenceExpert>> documentUploadEvidenceType,
                                   Optional<Bundle> bundleDetails, String docType) {
-        if (null != documentUploadEvidenceType) {
-            documentUploadEvidenceType.forEach(uploadEvidenceDocumentTypeElement -> {
-                if (uploadEvidenceDocumentTypeElement.getValue().getCreatedDatetime().isAfter(bundleDetails.get().getCreatedOn().orElse(
-                    null))) {
-                    uploadedEvidenceAfterBundle.add(ElementUtils.element(UploadEvidenceDocumentType.builder()
-                                                                             .typeOfDocument(docType)
-                                                                             .createdDatetime(
-                                                                                 uploadEvidenceDocumentTypeElement.getValue().getCreatedDatetime())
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/EvidenceUploadHandlerBase.java
```diff
@@ -355,57 +355,60 @@ SubmittedCallbackResponse buildConfirmation(CallbackParams callbackParams) {
...
-                                                                                 uploadEvidenceDocumentTypeElement.getValue().getExpertDocument())
-                                                                             .build()));
-                }
-            });
+        if (null == documentUploadEvidenceType) {
+            return;
         }
+        documentUploadEvidenceType.forEach(uploadEvidenceDocumentTypeElement -> {
+            if (uploadEvidenceDocumentTypeElement.getValue().getCreatedDatetime().isAfter(bundleDetails.get().getCreatedOn().orElse(
+                null))) {
```
filippo-possenti-version1, line 403: This also can be short-circuited, I believe

src/main/java/uk/gov/hmcts/reform/civil/handler/event/BundleCreationTriggerEventHandler.java
```diff
@@ -47,29 +47,31 @@ public class BundleCreationTriggerEventHandler {
...
+        coreCaseDataService.submitUpdate(caseId, caseContent);
     }
 
     boolean getIsBundleCreatedForHearingDate(Long caseId) {
         boolean isBundleCreated = false;
         CaseData caseData = caseDetailsConverter.toCaseData(coreCaseDataService.getCase(caseId).getData());
         List<IdValue<Bundle>> caseBundles = caseData.getCaseBundles();
-        isBundleCreated = !(caseBundles.stream().filter(bundleIdValue -> bundleIdValue.getValue()
+        isBundleCreated =
+            !(caseBundles.stream().filter(bundleIdValue -> bundleIdValue.getValue().getBundleHearingDate().isPresent()).filter(bundleIdValue -> bundleIdValue.getValue()
```
filippo-possenti-version1, line 74: This is different than before.
Before if there was an element in caseBundles then you would have extracted its bundleHearingDate.
Now you filter out the caseBundles that don't have a bundleHearingDate first and then on the remaining ones you run the check.
Right or wrong, this changes the behaviour of the application and we're late for that since everything has already been tested and approved.

I'm willing to sign off on this change as it's fairly small and in theory it won't happen for our bundles but next time we'll handle this situation as a separate bug or tech debt ticket.

Also, the condition is that tests are not altered. If this change resulted in the alteration of a test, then please revert it.

src/main/java/uk/gov/hmcts/reform/civil/helpers/BundleRequestMapper.java
```diff
@@ -152,18 +152,19 @@ private BundlingCaseData mapRespondent2Applicant2Details(BundlingCaseData bundli
 
     private ServedDocument mapServedDocuments(ServedDocumentFiles servedDocumentFiles) {
         List<BundlingRequestDocument> bundlingServedDocFiles = new ArrayList<>();
-        if (!Optional.ofNullable(servedDocumentFiles).isEmpty() && null != servedDocumentFiles.getParticularsOfClaimDocument()) {
-            servedDocumentFiles.getParticularsOfClaimDocument().forEach(document -> {
-                bundlingServedDocFiles.add(BundlingRequestDocument.builder()
+        if (null == servedDocumentFiles || null == servedDocumentFiles.getParticularsOfClaimDocument()) {
+            return null;
```
filippo-possenti-version1, line 156: Wrong return value.

If you want to short-circuit this, you need to make sure that it will return what the previous implementation was returning, which in this instance is never null. This needs to be as follows:

```suggestion
            return  ServedDocument.builder().particularsOfClaimDocument(ElementUtils.wrapElements(new ArrayList<>())).build();
```

src/main/java/uk/gov/hmcts/reform/civil/helpers/BundleRequestMapper.java
```diff
@@ -185,88 +186,101 @@ private List<Element<BundlingRequestDocument>> mapSystemGeneratedcaseDocument(Li
...
-            }
+        }
+        if (null != orderSDODocumentDJ) {
+            bundlingSystemGeneratedCaseDocs.add(BundlingRequestDocument.builder()
+                                                    .documentFileName(orderSDODocumentDJ.getDocumentFileName())
+                                                    .documentLink(DocumentLink.builder()
+                                                                      .documentUrl(orderSDODocumentDJ.getDocumentUrl())
+                                                                      .documentBinaryUrl(orderSDODocumentDJ.getDocumentBinaryUrl())
+                                                                      .documentFilename(orderSDODocumentDJ.getDocumentFileName()).build())
+                                                    .documentType(DocumentType.SDO_ORDER.name()).build());
```
filippo-possenti-version1, line 197: The "contract" of this ticket is simply to reduce complexity by leveraging short-circuiting, not altering the logic of the application.

Why has this been moved outside the "if" on line 174?

src/main/java/uk/gov/hmcts/reform/civil/helpers/BundleRequestMapper.java
```diff
@@ -185,88 +186,101 @@ private List<Element<BundlingRequestDocument>> mapSystemGeneratedcaseDocument(Li
...
-                bundlingWitnessDocs.add(BundlingRequestDocument.builder()
-                                            .documentFileName(fileNameBuilder.toString())
-                                            .documentLink(DocumentLink.builder()
-                                                              .documentUrl(document.getDocumentUrl())
-                                                              .documentBinaryUrl(document.getDocumentBinaryUrl())
-                                                              .documentFilename(document.getDocumentFileName()).build())
-                                            .build());
-
-            });
+        if (null == uploadEvidenceWitness || uploadEvidenceWitness.isEmpty()) {
```
filippo-possenti-version1, line 204: You don't need the `isEmpty()` check. If the collection is empty, no iteration will be ran.

Replace it as follows:
```suggestion
        if (null == uploadEvidenceWitness) {
```

src/main/java/uk/gov/hmcts/reform/civil/helpers/BundleRequestMapper.java
```diff
@@ -185,88 +186,101 @@ private List<Element<BundlingRequestDocument>> mapSystemGeneratedcaseDocument(Li
...
-                bundlingExpertDocs.add(BundlingRequestDocument.builder()
-                                           .documentFileName(fileNameBuilder.toString())
-                                           .documentLink(DocumentLink.builder()
-                                                             .documentUrl(document.getDocumentUrl())
-                                                             .documentBinaryUrl(document.getDocumentBinaryUrl())
-                                                             .documentFilename(document.getDocumentFileName()).build())
-                                           .build());
-
-            });
+        if (null == uploadEvidenceExpert || uploadEvidenceExpert.isEmpty()) {
```
filippo-possenti-version1, line 235: You don't need the `isEmpty()` check on uploadEvidenceExpert. If the collection is empty, the forEach just won't do anything.

Can be changed as follows:
```suggestion
        if (null == uploadEvidenceExpert) {
```

src/main/java/uk/gov/hmcts/reform/civil/helpers/BundleRequestMapper.java
```diff
@@ -185,88 +186,101 @@ private List<Element<BundlingRequestDocument>> mapSystemGeneratedcaseDocument(Li
...
-                bundlingExpertDocs.add(BundlingRequestDocument.builder()
-                                           .documentFileName(document.getDocumentFileName())
-                                           .documentLink(DocumentLink.builder()
-                                                             .documentUrl(document.getDocumentUrl())
-                                                             .documentBinaryUrl(document.getDocumentBinaryUrl())
-                                                             .documentFilename(document.getDocumentFileName()).build())
-                                           .build());
-
-            });
+        if (null == otherDocsEvidenceUpload || otherDocsEvidenceUpload.isEmpty()) {
```
filippo-possenti-version1, line 268: See above comments. Consider the following:

```suggestion
        if (null == otherDocsEvidenceUpload) {

```

src/test/java/uk/gov/hmcts/reform/civil/handler/callback/user/EvidenceUploadApplicantHandlerTest.java
```diff
@@ -604,10 +605,10 @@ void shouldBreakWhenThereIsAnyCaseBundlesWithNullCreatedDate() {
...
 
         // When: handler is called
-        // Then: an exception is thrown
-        assertThrows(NullPointerException.class, () -> {
-            handler.handle(params);
-        });
+        var response = (AboutToStartOrSubmitCallbackResponse) handler.handle(params);
+        CaseData updatedData = mapper.convertValue(response.getData(), CaseData.class);
+        // Then: applicant docs uploaded after bundle should return size 0
+        assertThat(updatedData.getApplicantDocsUploadedAfterBundle()).isNull();
```
filippo-possenti-version1, line 611: If you had to modify a test, it means that the behaviour has changed.

I seem to recall we added this test specifically.

Please don't alter the logic ad this point in time. The ticket has already been approved. If you alter the logic, test will have to be re-run and PO approval sought after again.

Restore this test and the related logic as before (except for any potential short-circuiting, which shouldn't affect the test anyway).

src/test/java/uk/gov/hmcts/reform/civil/handler/callback/user/EvidenceUploadRespondentHandlerTest.java
```diff
@@ -1190,10 +1192,10 @@ void shouldBreakWhenThereIsAnyCaseBundlesWithNullCreatedDate() {
...
 
         // When: handler is called
-        // Then: an exception is thrown
-        assertThrows(NullPointerException.class, () -> {
-            handler.handle(params);
-        });
+        var response = (AboutToStartOrSubmitCallbackResponse) handler.handle(params);
+        CaseData updatedData = mapper.convertValue(response.getData(), CaseData.class);
+        // Then: respondent docs uploaded after bundle should return size 0
+        assertThat(updatedData.getRespondentDocsUploadedAfterBundle()).isNull();
```
filippo-possenti-version1, line 1198: If you are changing the test, it means you are changing the logic.
We are not supposed to change the logic at this stage.

Please revert this test change and the changes that required this modification to the test.

filippo-possenti-version1 (changes requested):
Requires some changes. Please review comments.

src/main/java/uk/gov/hmcts/reform/civil/helpers/BundleRequestMapper.java
filippo-possenti-version1: This has been reverted and will be addressed in a separate ticket, should a problem arise.

PR #2417, opened 2023-03-09
CIV-7441 Move clients to commons module

sabahirfan:
CIV-7441

### Change description ###

This PR creates a commons-civil gradle module and moves the clients and relevant models into that module. 

Once all the common code is moved to this new module then we will extract out into it's own repo with pipelines. etc.

Will try to create small PRs to avoid conflicts with teams working in parallel.

**This is first of few PRs coming after this.** 

```
[ ] Yes
[x] No
```


asthamalviya (approved):

commons-civil/src/main/java/uk/gov/hmcts/reform/sendgrid/EmailAttachment.java
```diff
@@ -1,4 +1,4 @@
-package uk.gov.hmcts.reform.civil.sendgrid;
+package uk.gov.hmcts.reform.sendgrid;
```
filippo-possenti-version1, line 1: I'm not sure I understand the reason of promoting these outside of the "civil" package.

The components we are developing, after all, are still part of the "civil" project. Wouldn't this choice risk introducing package conflicts or confuse people as to where these classes live?

commons-civil/src/main/java/uk/gov/hmcts/reform/ccd/model/Address.java
```diff
@@ -0,0 +1,84 @@
...
+import uk.gov.hmcts.reform.civil.utils.ObjectUtils;
+
+import static uk.gov.hmcts.reform.civil.utils.StringUtils.joinNonNull;
+
+@Data
+@Builder(toBuilder = true)
+@JsonNaming(PropertyNamingStrategies.UpperCamelCaseStrategy.class)
+@NoArgsConstructor
+@AllArgsConstructor
+public class Address {
```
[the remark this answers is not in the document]
sabahirfan: This client was using the Address class from civil models. which doesn't seems right.

commons-civil/src/main/java/uk/gov/hmcts/reform/sendgrid/EmailAttachment.java
sabahirfan: As these clients are not specific to Civil projects. so kept these outside. The reusabsle components which are specific to Civil only are going to be inside that package.

asthamalviya (approved):

indu2019 (approved):

PR #2514, opened 2023-03-24
CIV-6668 Create message correlation with tenant-id 

jakiurmalik:
https://tools.hmcts.net/jira/browse/CIV-6668

### Change description ###

Emit Camunda messages with tenant id, if definition doesn't exist attempt without tenant id

Added new tests for changes

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/service/EventEmitterService.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/service/EventEmitterService.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.reform.civil.service;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.camunda.bpm.engine.RuntimeService;
+import org.camunda.bpm.extension.rest.exception.RemoteProcessEngineException;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/civil/service/EventEmitterService.java
```diff
@@ -23,20 +24,36 @@ public void emitBusinessProcessCamundaEvent(CaseData caseData, boolean dispatchP
...
                 applicationEventPublisher.publishEvent(new DispatchBusinessProcessEvent(caseId, businessProcess));
             }
-
-            log.info("Camunda event emitted successfully");
+            log.info("Camunda event emitted successfully with tenant");
         } catch (Exception ex) {
-            log.error(format("Emitting %s camunda event failed for case: %d, message: %s",
-                             camundaEvent, caseId, ex.getMessage()
-            ));
+            nullTenantAttempt = true;
```
sampankumar, line 39: Do we know any specific exception which is thrown in case tenant id attempts fails because of reason of not able to find the process and can we then catch that specific exception here?

PR #2555, opened 2023-04-04
CIV-7773 Fix fetch only 10 rows CUI dashboard

LeonardoPalmeiro:

https://tools.hmcts.net/jira/browse/CIV-7773

### Change description ###
Fix fetch just 10 rows

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/service/citizenui/DashboardClaimInfoService.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/service/citizenui/DashboardClaimInfoService.java
@@ -16,7 +16,6 @@ import uk.gov.hmcts.reform.civil.service.CoreCaseDataService;
 import uk.gov.hmcts.reform.civil.service.claimstore.ClaimStoreService;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/civil/service/citizenui/DashboardClaimInfoService.java
```diff
@@ -43,12 +45,22 @@ public List<DashboardClaimInfo> getClaimsForDefendant(String authorisation, Stri
...
     private List<DashboardClaimInfo> getCases(String authorisation) {
-        Query query = new Query(QueryBuilders.matchAllQuery(), emptyList(), 0);
-        SearchResult claims = coreCaseDataService.searchCases(query, authorisation);
-        if (claims.getTotal() == 0) {
-            return Collections.emptyList();
-        }
-        return translateSearchResultToDashboardItems(claims);
+        List<DashboardClaimInfo> dashboardClaimItems = new ArrayList<>();
+
+        int batchSize = BATCH_SIZE; // fetch 100 cases per batch
```
TatsyMaslova, line 50: is it approved by PO that we show all 100 cases in dashboard? what happens when we have more than 100 cases?

LeonardoPalmeiro: I changed any way because just doesn't work and we don't needed the BATCH_SIZE

PR #3168, opened 2023-08-29
CIV-10356 create api in civil service to return user caserole

LeonardoPalmeiro:

https://tools.hmcts.net/jira/browse/CIV-10356

### Change description ###
image

```
[ ] Yes
[x] No
```


m-meulendijks-v1 (dismissed):

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
```diff
@@ -204,4 +206,17 @@ public ResponseEntity<CaseDetails> caseworkerSubmitEvent(
         }
     }
 
+    @GetMapping(path = "/{caseId}/userCaseRoles")
+    @Operation(summary = "Get user Roles for a case")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = "OK"),
+        @ApiResponse(responseCode = "400", description = "Not Found"),
```
kurt-uk, line 213: Not found is 404

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
```diff
@@ -204,4 +206,17 @@ public ResponseEntity<CaseDetails> caseworkerSubmitEvent(
...
+        @ApiResponse(responseCode = "401", description = "Not Authorized")})
+    public ResponseEntity<List<String>> getUserInfo(
+        @PathVariable("caseId") String caseId,
+        @RequestHeader(HttpHeaders.AUTHORIZATION) String authorization) {
+        List<String> roles = userInformationService.getUserCaseRoles(caseId, authorization);
+        return new ResponseEntity<>(roles, HttpStatus.OK);
```
kurt-uk, line 219: can we place the call in the ok response body?
```suggestion
        return new ResponseEntity.ok(userInformationService.getUserCaseRoles(caseId, authorization));
```

src/main/java/uk/gov/hmcts/reform/civil/service/user/UserInformationService.java
```diff
@@ -0,0 +1,36 @@
...
+import uk.gov.hmcts.reform.civil.service.UserService;
+import uk.gov.hmcts.reform.idam.client.models.UserInfo;
+
+import java.util.List;
+
+@Service
+@RequiredArgsConstructor
+public class UserInformationService {
+
+    Logger log = LoggerFactory.getLogger(UserInformationService.class);
```
kurt-uk, line 18: using the standard config for logging, it would be easier to annotate the class with Slf4j

maciejmodzelewski (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/user/UserInformationService.java
```diff
@@ -0,0 +1,36 @@
...
+    public List<String> getUserCaseRoles(String caseId, String authorization) {
+        try {
+            UserInfo userInfo = userService.getUserInfo(authorization);
+            List<String> roles = coreCaseUserService.getUserCaseRoles(
+                caseId,
+                userInfo.getUid()
+            );
+            return roles;
+        } catch (Exception e) {
+            log.error(String.format("No case found for %s", caseId));
```
kurt-uk, line 32: you could also hit this catch if you try access a case that youre not on, make sure youre logging the error message here otherwise itll be really difficult to see when this goes wrong for the user not being on the case

kurt-uk (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/service/user/UserInformationService.java
kurt-uk: or if this is the camunda exception that doesnt provide much info, then maybe just tweak the error message to also account for user not being on case too

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
LeonardoPalmeiro: this was my initial thought but seems like civil-service doesn't like 404 :)

image

src/main/java/uk/gov/hmcts/reform/civil/service/user/UserInformationService.java
```diff
@@ -0,0 +1,36 @@
...
+    private final CoreCaseUserService coreCaseUserService;
+
+    public List<String> getUserCaseRoles(String caseId, String authorization) {
+        try {
+            UserInfo userInfo = userService.getUserInfo(authorization);
+            List<String> roles = coreCaseUserService.getUserCaseRoles(
+                caseId,
+                userInfo.getUid()
+            );
+            return roles;
```
kurt-uk, line 30: roles can be empty if the user id is wrong on the case, maybe return error for that too

kurt-uk (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/service/user/UserInformationService.java
LeonardoPalmeiro: what do you think 404 when we don't have role?

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
kurt-uk: hahaha just need to update the description then to bad request 😊

src/main/java/uk/gov/hmcts/reform/civil/service/user/UserInformationService.java
kurt-uk: makes sense to me!

src/test/java/uk/gov/hmcts/reform/civil/service/user/UserInformationServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public static final String CASE_ID = "123";
+    @Mock
+    private UserService userService;
+
+    @Mock
+    private CoreCaseUserService coreCaseUserService;
+
+    private UserInformationService userInformationService;
+
+    @BeforeEach
```
kurt-uk, line 38: try make this @BeforeAll to reduce stress on resources

kurt-uk (dismissed):

kurt-uk:
LGTM! 👍🏼

maciejmodzelewski (dismissed):

hmcts-version1-ignacio (dismissed):

kurt-uk (approved):

kurt-uk (approved):

maciejmodzelewski (approved):

PR #3174, opened 2023-08-31
CIV-10126 Set Up Lip Claim Issue After Payment

kenneth-hmcts:
https://tools.hmcts.net/jira/browse/CIV-10126

### Change description ###
Set Up LiP Claim Issue After Payment

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/CreateClaimLipCallBackHandler.java
```diff
@@ -66,6 +68,8 @@ private CallbackResponse lipClaimInitialState(CallbackParams callbackParams) {
     private CallbackResponse submitClaim(CallbackParams callbackParams) {
         CaseData.CaseDataBuilder caseDataBuilder = callbackParams.getCaseData().toBuilder();
         caseDataBuilder.submittedDate(time.now());
+        // Add back Pip in post to temporary pass the email event
+        caseDataBuilder.respondent1PinToPostLRspec(defendantPinToPostLRspecService.buildDefendantPinToPost());
```
TatsyMaslova, line 72: is this temporarily? because we need to add the pin only after the payment is successful. otherwise it will expire

neeta-hmcts: yes.. This is temporarily.. added because as we saw error while testing 'Respond To Claim' journey.
Later we may need to move this in new handler where we are adding payment service changes.

neeta-hmcts (dismissed):

sankhajuria (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/CreateClaimLipCallBackHandler.java
TatsyMaslova: ok but instead of adding this temporarily because we are actually developing journey after the claim has been successful we can generate pip when the claim is issued. no need to do anything temporarily. and according to the camunda we are going into the ClaimIssueForSpecCallbackHandler where we are adding a deadline for claimant. there we can add generating pin. or we could add a camunda process that would add a pin in the post. but i can see that for lr v lip we are adding the pin in post during creation of the claim which could have a bug - if the payment is delayed then the pin will expire. so i think it would be ok for now. as long as we agree with both cases (lr v lip and lip v lip) and not forget to update this

TatsyMaslova (dismissed):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/DefendantResponseApplicantForCuiNotificationHandler.java
```diff
@@ -60,17 +60,26 @@ public List<CaseEvent> handledEvents() {
```
neeta-hmcts, line 60: This File has changes for 'Respond To Claim' journey notification event update for LiP Claimant.

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/RepresentativeService.java
```diff
@@ -77,6 +77,9 @@ public Representative getRespondent2Representative(CaseData caseData) {
     }
```
neeta-hmcts, line 78: Fixed case document generation issue seen during 'Respond to Claim' for LiPvsLiP

src/main/resources/application.yaml
```diff
@@ -296,6 +296,9 @@ notifications:
```
neeta-hmcts, line 296: added 2 new notification for LiP Claimant

TatsyMaslova (approved):

portizdev (approved):

rajakm (approved):

PR #3209, opened 2023-09-08
CIV-6431 Debt Respite Translation

kenneth-hmcts:
https://tools.hmcts.net/jira/browse/CIV-6431

### Change description ###
Enter Breathing Space Translation to ccd

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/EnterBreathingSpaceCuiCallbackHandler.java
```diff
@@ -0,0 +1,56 @@
+package uk.gov.hmcts.reform.civil.handler.callback.user;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
```
TatsyMaslova, line 3: I wonder whether instead of calling it CUI we could call it LIP? because in theory that handler could be called from some other program. but it is to do with breathing space in LIP case. what do you think?

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/EnterBreathingSpaceCuiCallbackHandler.java
```diff
@@ -0,0 +1,56 @@
...
+    private final ObjectMapper objectMapper;
+
+    @Override
+    public List<CaseEvent> handledEvents() {
+        return EVENTS;
+    }
+
+    @Override
+    protected Map<String, Callback> callbacks() {
+        return Map.of(
```
TatsyMaslova, line 39: could have a map as a final field. so that every time we call a method it does not create a new map

src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowStateAllowedEventService.java
```diff
@@ -788,6 +789,7 @@ public class FlowStateAllowedEventService {
             CLAIM_ISSUED_PAYMENT_FAILED.fullName(),
             List.of(
                 ENTER_BREATHING_SPACE_SPEC,
+                ENTER_BREATHING_SPACE_CUI,
```
TatsyMaslova, line 792: ENTER_BREATHING_SPACE_LIP?

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/EnterBreathingSpaceCuiCallbackHandler.java
kenneth-hmcts: Updated to LIP instead using CUI

src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowStateAllowedEventService.java
kenneth-hmcts: Updated

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/EnterBreathingSpaceCuiCallbackHandler.java
kenneth-hmcts: For this one, I have tried to change but not successful, there is error after changing protected to final. Also, I have checked other handler and found that they are similar to the one that I am using

TatsyMaslova: like this Kenneth for example: 
image

TatsyMaslova (approved):

neeta-hmcts (approved):

portizdev (approved):

rajakm (approved):

PR #3258, opened 2023-09-20
CIV-000 Add missing config for staging

sankaviv1:
*Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/civil-service/values.aat.template.yaml
+++ b/charts/civil-service/values.aat.template.yaml
@@ -42,6 +42,7 @@ java:
     DOCUMENT_MANAGEMENT_SECURED: true
     STITCHING_API_ENABLED: true
     RELEASE_REVISION: "{{ .Release.Revision }}"
+    ACA_SERVICE_API_BASEURL: http://${SERVICE_NAME}-aac-manage-case-assignment
     HMC_HEARINGS_SUBSCRIPTION_ENABLED: false
   postgresql:
     enabled: true


charts/civil-service/values.aat.template.yaml
```diff
@@ -42,6 +42,7 @@ java:
     DOCUMENT_MANAGEMENT_SECURED: true
     STITCHING_API_ENABLED: true
     RELEASE_REVISION: "{{ .Release.Revision }}"
+    ACA_SERVICE_API_BASEURL: http://${SERVICE_NAME}-aac-manage-case-assignment
```
vasudevganesanhmcts, line 45: @sankaviv1  will this fix the problem fully? as the error was the api tests trying to access AAC_API_URL which is set as https://manage-case-assignment-civil-service-staging-aat.service.core-compute-aat.internal. .which is also the ingress host set on the pod ? is there any civil ccd master ui tests of NOC which passed correctly ? as this may be due to load balancer issue config not set

sankaviv1: master ccd is running the e2e tests for NoC @vasudevganesanhmcts 

this one line was the difference between the config defined in this repo for staging and preview (and these tests were previously running on preview)

vasudevganesanhmcts: Yes i checked the tests . .but that test doesnt have requests directly to case assignment service restHelper.request(
      `${config.url.caseAssignmentService}/noc/noc-questions?case_id=${caseId}`, headers, null, 'GET') 
anyway just my thoughts . .  happy to approve this if you feel this fixes it

HarryH96: LGTM and observe the behaviour - we wont know as staging configuration is very fiddly.
We have ensured that the civil-service has been whitelisted on AAT under this PR:
https://github.com/hmcts/cnp-flux-config/pull/24963

PR #3265, opened 2023-09-21
CIV-10659 - ensure null pointer exception is not thrown when repaymen…

TatsyMaslova:

https://tools.hmcts.net/jira/browse/CIV-10659

### Change description ###

```
[ ] Yes
[ ] No
```


neeta-hmcts (approved):

sankhajuria (commented):
Have you checked why we are getting null object ?

TatsyMaslova:
> Have you checked why we are getting null object ?

yes, only happens with Raja's test script - it has a fault where it puts faulty values that as a result cannot be saved to ccd. so repayment plan field is null

rajakm (approved):

PR #3286, opened 2023-09-25
CIV-10753 Added location whitelisting mechanism

Gareth40343:
https://tools.hmcts.net/jira/browse/CIV-10753

- Added location whitelisting mechanism.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/FeatureToggleService.java
```diff
@@ -71,4 +71,9 @@ public boolean isUpdateContactDetailsEnabled() {
     public boolean isLipVLipEnabled() {
         return featureToggleApi.isFeatureEnabled("cuiReleaseTwoEnabled");
     }
+
+    public boolean isLocationWhiteListedForCaseProgression(String locationEpimms) {
+        return featureToggleApi.isFeatureEnabledForLocation("case-progression-location-whitelist", locationEpimms,
+                                                            true);
```
sherlynkhaw, line 77: not sure if the default value should be true/false?

if LD is down or if the whitelist is removed, what are we expecting to see? all users or no users accessing?

Gareth40343: Set the default to false to be inline with other toggles :)

sankaviv1 (approved):

sherlynkhaw (approved):

PR #3297, opened 2023-09-26
CIV-9616 Defendant Latest Update Mediation Succesful PDF

jeswanth-hmcts:

https://tools.hmcts.net/jira/browse/CIV-9616

### Change description ###

Defendant Latest Update Mediation Succesful PDF

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/MediationSuccessfulCallbackHandler.java
```diff
@@ -41,12 +49,47 @@ public List<CaseEvent> handledEvents() {
...
             .data(caseDataUpdated.toMap(objectMapper))
             .state(CASE_STAYED.name())
             .build();
     }
+
+    private List<Element<ManageDocument>> updateManageDocumentsListWithMediationAgreementDocument(
+        CallbackParams callbackParams) {
+        CaseData caseData = callbackParams.getCaseData();
+        List<Element<ManageDocument>> manageDocumentsList =
+            Optional.ofNullable(caseData.getManageDocuments()).orElse(new ArrayList<>());
```
TatsyMaslova, line 70: we are repeating code here and in the further up. perhaps this knowledge should be inside casedata. then we will just have to call that method on case data here and elsewhere where we would require it. In addition we could unit test this method on case data test

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/MediationSuccessfulCallbackHandler.java
```diff
@@ -41,12 +49,47 @@ public List<CaseEvent> handledEvents() {
...
+                .documentLink(document.getDocument())
+                .documentName(document.getName())
+                .documentType(mapDocumentTypeToManageDocumentType(document.getDocumentType()))
+                .build();
+            manageDocumentsList.add(element(manageDocument));
+        });
+        return manageDocumentsList;
+    }
+
+    private ManageDocumentType mapDocumentTypeToManageDocumentType(DocumentType documentType) {
```
TatsyMaslova, line 87: this mapping should not be inside the handler. what if this knowledge is needed else where? hanlder should not have a lot of business logic. think about it as a controller. perhaps the mapping should happen in the ManageDocumentType enum?

jeswanth-hmcts: moved to a util class

src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -982,13 +982,19 @@ public List<Element<RecurringIncomeLRspec>> getRecurringIncomeForRespondent1() {
...
+                .orElse(
+                    null);
         }
         return Optional.ofNullable(getRespondent1DQ()).map(Respondent1DQ::getRespondent1DQRecurringExpenses).orElse(
             null);
     }
 
+    @JsonIgnore
+    public List<Element<ManageDocument>> getManageDocumentsList() {
+        return Optional.ofNullable(getManageDocuments()).orElse(new ArrayList<>());
```
TatsyMaslova, line 995: Collections.emptyList() would be better

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/MediationSuccessfulCallbackHandler.java
```diff
@@ -41,12 +47,37 @@ public List<CaseEvent> handledEvents() {
...
             .data(caseDataUpdated.toMap(objectMapper))
             .state(CASE_STAYED.name())
             .build();
     }
+
+    private List<Element<ManageDocument>> updateManageDocumentsListWithMediationAgreementDocument(
+        CallbackParams callbackParams) {
+        CaseData caseData = callbackParams.getCaseData();
+        List<Element<ManageDocument>> manageDocumentsList = caseData.getManageDocumentsList();
+        Optional<MediationAgreementDocument> mediationAgreementDocument = Optional.ofNullable(caseData)
```
TatsyMaslova, line 68: how come caseData could be null? perhaps it should be Optional.ofNullable(casdata.getMediation()).map(Mediation::getMediationSuccessful).map(MediationSuccssful::getMediationAgreement)

laperla11 (approved):

TatsyMaslova (approved):

ElifDenizEsea (approved):

neeta-hmcts (approved):

rajakm (approved):

PR #3305, opened 2023-09-27
CIV-0000 fix issue with lr v lr sealed response generator

TatsyMaslova:

N/A

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/sealedclaim/SealedClaimResponseFormGeneratorForSpec.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/sealedclaim/SealedClaimResponseFormGeneratorForSpec.java
@@ -39,6 +39,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Stream;
 
+import static uk.gov.hmcts.reform.civil.enums.MultiPartyScenario.isOneVOne;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/sealedclaim/SealedClaimResponseFormGeneratorForSpec.java
```diff
@@ -141,6 +140,12 @@ public SealedClaimResponseFormForSpec getTemplateData(CaseData caseData, String
         return builder.build();
     }
 
+    private void addRepaymentPlanDetails(SealedClaimResponseFormForSpec.SealedClaimResponseFormForSpecBuilder builder, CaseData caseData) {
+        if (featureToggleService.isLipVLipEnabled() && caseData.getRespondent2() == null) {
```
sampankumar: If its multiparty scenario but respondent 1 has responded then do we want this code to run?

TatsyMaslova: no probably not. you are right. i am going to add the single one v one check instead

neeta-hmcts (approved):

PR #3329, opened 2023-10-02
CIV-10858 fasttrack hearing fee

manish14836:

https://tools.hmcts.net/jira/browse/CIV-10858

### Change description ###
Added functionality to get the hearing fee for fast track claims from Fee api

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/fees/client/FeesClient.java
```diff
@@ -27,22 +29,45 @@ public FeesClient(
...
-        String keyword = event.equalsIgnoreCase("issue")
-            ? "MoneyClaim"
-            : "HearingSmallClaims";
-
         if (featureToggleService.isFeatureEnabled("fee-keywords-enable")) {
-            return this.feesApi.lookupFee(service, jurisdiction1, jurisdiction2, channel, event, keyword, amount);
+            String keyword;
+            String jurisdiction2;
+
+            if (featureToggleService.isLipVLipEnabled() && "hearing".equalsIgnoreCase(event) && AllocatedTrack.FAST_CLAIM == AllocatedTrack.getAllocatedTrack(
```
jarekPierchala, line 48: Put that condition in If statement to separate method and use the method in If ;)

jarekPierchala (changes requested):

src/main/java/uk/gov/hmcts/reform/fees/client/FeesClient.java
TatsyMaslova, line 48: perhaps instead of having these as hardcode strings here we could put them in constants? i mean "hearing, "FastTrackHrg", "issue",   "MoneyClaim", and "HearingSmallClaims". or you can go even further and create Event enum and hearingKey enum. then we can use those in test too so that in the instance when the key is changed we change it only in one place for example

TatsyMaslova (approved):

neeta-hmcts (dismissed):

TatsyMaslova (approved):

jarekPierchala (approved):

PR #3373, opened 2023-10-09
CIV-8926 repayment calculator

TatsyMaslova:

https://tools.hmcts.net/jira/browse/CIV-8926

### Change description ###

create court decision calculator for claimant proposed repayment plan

```
[ ] Yes
[ ] No
```


ElifDenizEsea (dismissed):

kenneth-hmcts (approved):

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
```diff
@@ -256,4 +260,14 @@ public ResponseEntity<List<String>> getUserInfo(
         return ResponseEntity.ok(userInformationService.getUserCaseRoles(caseId, authorization));
     }
 
+    @PostMapping(path = "/{caseId}/courtDecision")
+    @Operation(summary = "Get decision on proposed claimant repayment")
+    public RepaymentDecisionType getDecisionOnClaimantProposedRepayment(
```
LeonardoPalmeiro, line 265: The method name getDecisionOnClaimantProposedRepayment suggests that it's a getter method, but it's actually performing an action. Consider renaming it to something like calculateDecisionOnClaimantProposedRepayment to better reflect its purpose.

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
```diff
@@ -256,4 +260,14 @@ public ResponseEntity<List<String>> getUserInfo(
...
+        @PathVariable("caseId") Long caseId,
+        @RequestHeader(HttpHeaders.AUTHORIZATION) String authorisation,
+        @RequestBody ClaimantProposedPlan claimantProposedPlan) {
+        var caseDetails = coreCaseDataService.getCase(caseId, authorisation);
+        return repaymentPlanDecisionService.getCalculatedDecision(caseDetails, claimantProposedPlan);
```
LeonardoPalmeiro, line 270: we could wrap on ResponseEntity
return ResponseEntity.ok(decision);

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
```diff
@@ -256,4 +260,14 @@ public ResponseEntity<List<String>> getUserInfo(
         return ResponseEntity.ok(userInformationService.getUserCaseRoles(caseId, authorization));
     }
 
+    @PostMapping(path = "/{caseId}/courtDecision")
+    @Operation(summary = "Get decision on proposed claimant repayment")
+    public RepaymentDecisionType getDecisionOnClaimantProposedRepayment(
+        @PathVariable("caseId") Long caseId,
+        @RequestHeader(HttpHeaders.AUTHORIZATION) String authorisation,
```
LeonardoPalmeiro, line 267: Follow consistent naming conventions for variables. For example, you can use authorization instead of authorisation for the header parameter.

LeonardoPalmeiro (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
TatsyMaslova: I dont think its a good approach Leo. it will return ok anyway if it is successful. we are returning a dto (json) here. have a look at other post methods

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
TatsyMaslova: if you have a look further in the controller we use authorisation here too. i prefer the british way of spelling it 
`  public Boolean caseworkerSearchCase(
        @PathVariable("userId") String userId,
        @RequestHeader(HttpHeaders.AUTHORIZATION) String authorization,
        @RequestParam(name = "sdtRequestId") String searchParam
    ) {
        CaseWorkerSearchCaseParams params = CaseWorkerSearchCaseParams.builder()
            .authorisation(authorization)
            .userId(userId)
            .searchCriteria(Map.of("case.sdtRequestIdFromSdt", searchParam)).build();
        List<CaseDetails> caseDetails = caseSdtRequestSearchService.searchCaseForSdtRequest(params);

        if (caseDetails.size() < 1 && caseDetails.isEmpty()) {
            return true;
        }
        return false;
    }
`

src/main/java/uk/gov/hmcts/reform/civil/controllers/cases/CasesController.java
TatsyMaslova: What i mean is we are implicitly returning a responsebody here (because it is a rest controller). so the response will be serialized into the body of the HTTP response. hence we dont need a response entity here

LeonardoPalmeiro (approved):

neeta-hmcts (approved):

neeta-hmcts (approved):

ElifDenizEsea (approved):

rajakm (approved):

PR #3390, opened 2023-10-11
CIV-0000 Prod LIP cases do not send event if email id is null

asthamalviya:
CIV-0000 Prod LIP cases do not send event if email id is null


indu2019 (approved):

MMNycz (approved):

HarryH96 (approved):
Approved on the basis that this is actually tackled properly - this is a temporary workaround to reduce pressure of stuck claims

MMNycz (approved):

zikrurR (approved):

PR #3617, opened 2023-11-22
CIV-11647 Claimant list matcher

ElifDenizEsea:

- [x] commit messages are meaningful and follow good commit message guidelines
- [x] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)

https://tools.hmcts.net/jira/browse/CIV-11647

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/civil/model/citizenui/DashboardClaimantClaimStatusFactory.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmcts.reform.civil.model.citizenui;
```
paul-pearson, line 1: There is no need for this class.  You can use the existing DashboardClaimStatusFactory because it is doing the same thing.  You will just supply it with the alternative matcher.

src/main/java/uk/gov/hmcts/reform/civil/service/citizenui/DashboardClaimInfoService.java
```diff
@@ -170,4 +178,16 @@ private int getCcdClaimsCount(final SearchResult ccdClaims) {
...
         return Optional.ofNullable(ccdClaims).map(SearchResult::getTotal).orElse(0);
     }
+
+    private DashboardClaimStatus getStatus(boolean isClaimant, CaseData caseData) {
+        return isClaimant
+            ? dashboardClaimantClaimStatusFactory.getDashboardClaimStatus(new CcdDashboardClaimantClaimMatcher(
+            caseData,
+            featureToggleService
+        ))
+            : dashboardClaimStatusFactory.getDashboardClaimStatus(new CcdDashboardClaimMatcher(
```
paul-pearson, line 188: Maybe the existing claim matcher could be renamed to CcdDashboardDefendantClaimMatcher if it is only to be used for defendant.

paul-pearson (changes requested):

src/main/java/uk/gov/hmcts/reform/civil/service/citizenui/DashboardClaimInfoService.java
ElifDenizEsea: done

src/main/java/uk/gov/hmcts/reform/civil/model/citizenui/DashboardClaimantClaimStatusFactory.java
paul-pearson: DashboardClaimStatusFactory can stay exactly the same as it was.  There is no need to have two different ClaimStatusFactories, because they are both doing the same thing.  It is only the matcher that we need to have two of.

jarekPierchala (approved):

build.gradle
```diff
@@ -315,7 +315,7 @@ sonarqube {
     property "sonar.projectKey", "civil-service"
     property "sonar.coverage.jacoco.xmlReportPaths", "${jacocoTestReport.reports.xml.destination.path}"
     property "sonar.coverage.exclusions", "**/model/**, **/config/**/*Configuration.java, **/testingsupport/**, **/*ExternalTaskListener.java, **/*BaseExternalTaskHandler.java, **/stereotypes/**, **/*Exception.java, **/EventHistoryMapper*.java, **/model/hearingvalues/**, **/enums/hearing/**, **/fees/client/**, **/enums/sdo/**, **/service/PaymentsService.java"
-    property "sonar.cpd.exclusions", "**/*DocumentManagementService.java, **/*Spec*.java"
+    property "sonar.cpd.exclusions", "**/*DocumentManagementService.java, **/*Spec*.java, **/*CcdDashboardClaimantClaimMatcher.java"
```
paul-pearson, line 318: Do we need to add  to sonar exclusions?  Currently, CcdDashboardClaimMatcher is tested via the CcdClaimStatusDashboardFactoryTest.  We should add new tests here using the new matcher as well.  For now they can match the existing tests as the implementation for both matchers is the same.  But this means we won't forget to test when we make changes to the new matcher.

paul-pearson (changes requested):

build.gradle
paul-pearson: @ElifDenizEsea has said that the exclusion is due to duplication rather than code coverage.  It would be worth moving the common implementations to a common base class, and then then we'd only need a test for each matcher separately when the implementation differs.  This could be done in one of the subsequent tickets if you prefer.

paul-pearson (approved):

kenneth-hmcts (approved):

jarekPierchala (approved):

portizdev (approved):

PR #3900, opened 2024-01-16
CIV-12060 Prevent creation of Default Judgment Spec form in LR V LiP claims

JackDevN:

https://tools.hmcts.net/jira/browse/CIV-12060

### Change description ###

This change introduces a check to see if the claim is LRvLiP to see if the DJ document should be generated or not.

```
[ ] Yes
[ ] No
```


asthamalviya (approved):

dtortolaV1 (dismissed):

German-A-Amoros (approved):

UshaPanneerselvam1 (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/docmosis/GenerateDJFormHandler.java
```diff
@@ -59,16 +59,18 @@ private CallbackResponse generateClaimForm(CallbackParams callbackParams) {
         CaseData caseData = callbackParams.getCaseData();
         CaseData.CaseDataBuilder caseDataBuilder = caseData.toBuilder();
 
-        if (ofNullable(caseData.getRespondent2()).isPresent()
-            && ((ofNullable(caseData.getDefendantDetails()).isPresent()
-            && caseData.getDefendantDetails().getValue().getLabel().startsWith(
+        if (!(caseData.isLRvLipOneVOne() && isSpecHandler(callbackParams))) {
+            if (ofNullable(caseData.getRespondent2()).isPresent()
+                && ((ofNullable(caseData.getDefendantDetails()).isPresent()
+                && caseData.getDefendantDetails().getValue().getLabel().startsWith(
```
dtortolaV1, line 65: instead of ((ofNullable(caseData.getDefendantDetails()).isPresent() && caseData.getDefendantDetails().getValue().getLabel().startsWith("Both")) you can use ofNullable(caseData.defendantDetails).map(className::getValue).map(className::getLabel).filter(s -> s.startsWith("Both"))
And the same with defendantDetailsSpec
This reduces in two the boolean operators (reducing the flow branches) and it is null safe if getValue or getLabel returns null (which shouldn't anyway)
Apart from that, ofNullable(something).isPresent() does the same as something != null. If you are not going to use "something" afterwards, I think it's more readable using != null, but your choice.
And down in line 70, ofNullable(caseData.getRespondent2()).isEmpty() is the same as caseData.getRespondent() == null
Last, since caseData.getRespondent2() appears in both conditions (once with isEmpty, the other with isPresent), you can change the order of the blocks:
now you have if (respondent2 != null && restOfCondition) { blockA } else if (respondent2 == null) {blockB}
you can do if (respondent2 == null) {blockB} else if (restOfCondition) {blockA}
This way you remove another boolean operator

dtortolaV1 (dismissed):

asthamalviya (approved):

UshaPanneerselvam1 (approved):

PR #4031, opened 2024-02-12
Adding Integration Tests

kannan-v-hmcts:

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -233,6 +233,10 @@ test {
   failFast = true
 }
 
+tasks.withType(Copy) {
+  duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+}
+
 task integration(type: Test) {
   description = "Runs integration tests"
   group = "Verification"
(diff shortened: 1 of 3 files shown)


laarmada (approved):
I don't like the name, because you are deleting and inserting no getting but good job, only my obsession :)

kannan-v-hmcts:
> I don't like the name, because you are deleting and inserting no getting but good job, only my obsession :)

This is not final. I don't like the name of the SQL file either and we can have a generic name so that it populates data for all our endpoints. The problem is that the other Integration tests that are there in DashboardControllerTest.java file is mocking the call the repository which I would like to discuss with the team and If we are mocking the database calls in the Integration Test, it beats the whole purpose of the Integration Tests. With Deleting and adding the data again in the SQL file, it complains that the data already exists even If I put in a new data with a completely new Id which is quite strange. As a workaround, I am deleting the data for now and adding. This needs to be looked at and that's why I just created a draft PR just for us to have an initial look

laarmada (approved):

m-meulendijks-v1 (approved):

dharmendrak (approved):

PR #4048, opened 2024-02-15
CIV-12627 : Adding scripts to update rolecategory for transfer online…

asthamalviya:
https://tools.hmcts.net/jira/browse/CIV-12627

DMN is fixed but for the cases where role category is not populated we are manually updating them as 'ADMIN' .
 With this PR I am adding the script which is going to run and fix that.

This change would be done under CHG5015221


src/main/resources/sqlscripts/CHG5015221_updating.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+update cft_task_db.tasks set role_category = 'ADMIN' where role_category is null and jurisdiction = 'CIVIL' and task_name in ('Request for transfer online case');
```
raghera, line 3: Are you planning to run any pre-checks before running this?  
How do you know if the records you expect to revert in the rollback script are the ones you've modified in this line?  You may want to capture the latest numbers just before running the script.  E.g. count total nulls, retrieve all the task_ids etc?

As an example: https://github.com/hmcts/wa-task-management-api/blob/master/src/main/resources/scripts/rwa-2905_update_civil_role_categories.sql

raghera: You may also want to run some post run verifications

asthamalviya: Hi Ravi , for pre check I have mentioned that in the CR too I will add those also as script in this PR so that you are assured.

asthamalviya: I have updated the folder with similar to attached script. Thank you

raghera: I don't see the scripts you mention above?

raghera (approved):
Looks ok as I assume whoever is running the steps in prod (from PlatOps) is guided in running the appropriate steps to perform.

PR #4092, opened 2024-02-23
CIV-12707 bilingual more info notification

ElifDenizEsea:

- [x] commit messages are meaningful and follow good commit message guidelines
- [x] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)

https://tools.hmcts.net/jira/browse/CIV-12707

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -367,6 +367,7 @@ notifications:
   notifyApplicantForHwFMoreInformationNeededWelsh: "65b3524b-b58c-4cdb-b8b5-5fc2cec505a5"
   notifyApplicantForHwfUpdateRefNumber: "7f87e629-6473-4732-8d80-4174d06696d5"
   notifyApplicantForHwfPartialRemission: "2c5c5cae-f1b1-41b5-afe5-0ce2e68f8d3c"
+  notifyApplicantForHwfPartialRemissionBilingual: "5bbd4518-5501-4737-8a1e-a6adacd5e679"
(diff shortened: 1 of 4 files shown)


src/main/resources/application.yaml
```diff
@@ -364,8 +364,11 @@ notifications:
   notifyApplicantForHwfNoRemission: "3008a18a-07ec-445b-a226-62ffa19132ce"
   notifyApplicantForHwfNoRemissionWelsh: "f9d2f0db-2ae4-4691-8076-bfbb82548846"
   notifyApplicantForHwFMoreInformationNeeded: "02956398-fffb-4461-93a0-3713788363a8"
+  notifyApplicantForHwFMoreInformationNeededWelsh: "65b3524b-b58c-4cdb-b8b5-5fc2cec505a5"
   notifyApplicantForHwfUpdateRefNumber: "7f87e629-6473-4732-8d80-4174d06696d5"
+  notifyApplicantForHwfUpdateRefNumberBilingual: "7f87e629-6473-4732-8d80-4174d06696d5"
```
paul-pearson, line 369: This property has been added but I don't think is currently being used.  Also it would be good to be consistent with the naming - we should either use "...Bilingual" or "...Welsh" but not both.

paul-pearson (changes requested):

src/main/resources/application.yaml
ElifDenizEsea: It is added to civil-commons for another ticket, so I needed to ad for tests to run.
for naming, I agree we should align on one of them

paul-pearson (approved):

PR #4106, opened 2024-02-27
CIV-12869 Ambiguous  error for state CLAIM_SUBMITTED

kalachandrasekar1:

https://tools.hmcts.net/jira/browse/CIV-12869

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowLipPredicate.java
```diff
@@ -11,7 +11,7 @@ private FlowLipPredicate() {
     }
 
     public static final Predicate<CaseData> isLipCase = caseData ->
-        caseData.isApplicantNotRepresented() || caseData.isApplicantRepresented();
+        caseData.isApplicantNotRepresented();
```
portizdev, line 14: Should we also check that the defendant is LiP? Maybe it would be better to use the isLipvLipOneVOnede caseData method.

jeswanth-hmcts (approved):

PR #4410, opened 2024-04-03
CIV-11302 Trial Arrangements Notify Other Party

kannan-v-hmcts:
https://tools.hmcts.net/jira/browse/CIV-11302


dashboard-notifications/src/main/java/uk/gov/hmcts/reform/dashboard/services/DashboardNotificationService.java
```diff
@@ -80,9 +80,14 @@ public void recordClick(UUID id, String authToken) {
...
                 .createdBy(idamApi.retrieveUserDetails(authToken).getFullName())
+                .createdAt(OffsetDateTime.now())
                 .build();
-
-            notificationActionRepository.save(notificationAction);
+            if (nonNull(notification.getNotificationAction())
+                && notification.getNotificationAction().getActionPerformed().equals("Click")) {
+                notificationAction.setId(notification.getNotificationAction().getId());
+            }
+            notification.setNotificationAction(notificationAction);
```
m-meulendijks-v1, line 89: So this is a slightly different way of saving it - did you discuss this with Dharmendra?

kannan-v-hmcts: Earlier the code was just adding new entries to the table but now that we are also updating the existing row if an entry is present already, if we use the notification action repository, it is throwing an error Multiple representations of the same entity [[uk.gov](http://uk.gov/).hmcts.reform.dashboard.entities.NotificationActionEntity#49] are being merged.  The reason being, we are fetching the data using the notification repository and directly updating using action repository. I think hibernate maintains a state using hashcode internally. Now we are saving it using the notification repository and it cascades/updates the action repository

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/dashboardnotifications/claimant/TrialArrangementsNotifyOtherPartyNotificationHandler.java
```diff
@@ -0,0 +1,44 @@
...
+        return TASK_ID;
+    }
+
+    @Override
+    public List<CaseEvent> handledEvents() {
+        return EVENTS;
+    }
+
+    @Override
+    public String getScenario(CaseData caseData) {
```
m-meulendijks-v1, line 41: Do we need a check to see if the claimant is not represented? Otherwise we are saving notifications in the database that will never be seen. DashboardCallbackHandler does have a method that should allow you to present a condition: 
```
    @Override
    public boolean shouldRecordScenario(CaseData caseData) {
        return caseData.isRespondent1NotRepresented();
    }
```

kannan-v-hmcts: Good point @m-meulendijks-v1 . Updated now

m-meulendijks-v1 (approved):

laarmada (approved):

PR #4411, opened 2024-04-03
CIV-11300 Hearing Scheduled notification

miguelMolina3691:

https://tools.hmcts.net/jira/browse/CIV-11300

### Change description ###

```
[ ] Yes
[ x ] No
```


dashboard-notifications/src/main/java/uk/gov/hmcts/reform/dashboard/services/DashboardScenariosService.java
kannan-v-hmcts, line 1: @miguelMolina3691 I don't think these changes are needed.  Notification and NotificationAction entities are having one to one relationship and ideally there should be only one entry in the notification action table for a notification. When I worked on my ticket, I noticed that each time the link is clicked,  a new entry is made in the table rather than updating the same row. I have a fix in my PR that updates the record if present or adds a  new entry if not found. https://github.com/hmcts/civil-service/pull/4410

PR #4437, opened 2024-04-08
DTSCCI-0000 version camunda deps together

linusnorton:
Note that the camunda starter project has moved to a community module and the application.yml feign config is slightly different: https://camunda-community-hub.github.io/camunda-platform-7-rest-client-spring-boot/snapshot/getting-started.html

I've switched to the recommended approach of using the default url to specify the camunda rest urls.


rishikrsharma (approved):
LGTM

PR #4546, opened 2024-04-19
CIV-11355 Trial Arrangements TaskList

kannan-v-hmcts:
https://tools.hmcts.net/jira/browse/CIV-11355


src/main/resources/db/migration/V2024_04_04_1632__CIV-11303_bundle_created_scenario.sql
```diff
@@ -35,4 +41,16 @@ values ('<a href={VIEW_BUNDLE} class="govuk-link">View the bundle</a>',
        ('<a href={VIEW_BUNDLE} class="govuk-link">View the bundle</a>',
         'Hearings',
         '<a href={VIEW_BUNDLE} class="govuk-link">View the bundle</a>',
-        'Hearings', 'Hearing.Bundle.View', 'Scenario.AAA6.CP.Bundle.Ready.Defendant', '{3, 3}', 'DEFENDANT', 8);
+        'Hearings', 'Hearing.Bundle.View', 'Scenario.AAA6.CP.Bundle.Ready.Defendant', '{3, 3}', 'DEFENDANT', 8),
+       ('<a href={VIEW_BUNDLE} class="govuk-link">View the bundle</a>',
+        'Hearings',
```
laarmada, line 46: Only a minor comment, I prefer to split the file between claimant and defendant if we modify one of them we can see in the flyway schema exactly which notification was changed. On the other hand we have splitted other ones.

laarmada (approved):

laarmada:
we use the same camunda for that 4 Scenarios cover in this issue? If make the same add and remove notification .... why we need to duplicate the values? Is possible to use the same event and we can be more efficient?

src/main/resources/db/migration/V2024_04_04_1632__CIV-11303_bundle_created_scenario.sql
kannan-v-hmcts: Hi @laarmada It was already in a single file and I just modified it. When we go to production, it will be a one time entry. We cannot modify the same file again once it is applied to the DB as flyway complains if it is modified. Any further changes has to be done through a new alter script in a new file. I am happy to split it into 2 separate files but I don't see any advantage really. @m-meulendijks-v1 did the original implementation. Your thoughts on this Max ?

kannan-v-hmcts:
> we use the same camunda for that 4 Scenarios cover in this issue? If make the same add and remove notification .... why we need to duplicate the values? Is possible to use the same event and we can be more efficient?

@laarmada We use 2 camunda tasks and each handles 2 scenarios (when the Trail arrangement is completed and when it is not). The task list needs to be updated based on the completion of the trial arrangement and thats why we are using 2 scenarios for claimant and 2 for defendant. I hope this answers your question. If not, I am happy to explain it over a call

src/main/resources/db/migration/V2024_04_04_1632__CIV-11303_bundle_created_scenario.sql
m-meulendijks-v1: I think migrations should be ticket based, we split it up between claimant/defendant before because that's how the tickets were split. I see no need to split this one.

dtortolaV1 (approved):

m-meulendijks-v1 (dismissed):

miguelMolina3691 (dismissed):

laarmada (approved):

laarmada (approved):

laarmada (approved):

miguelMolina3691 (approved):

PR #4632, opened 2024-04-30
CIV-13467 bug

laarmada:

[CIV-13467](https://tools.hmcts.net/jira/browse/CIV-13467)
### Change description ###
remove the "Add the trial arrangements" from other not related scenarios "Task list items only"

```
[ ] Yes
[x] No
```


dharmendrak (dismissed):

m-meulendijks-v1 (commented):
Shouldn't the approach be to have the Trial Arrangements task item be visible from the beginning for fast track, and absent from the beginning for small claims?

If I am reading the sql changes correctly, the changes in this PR are removing the task item from both fast track & small claims, only for it to be added to fast track at the Trial Arrangements required stage.

So we need to differentiate between Fast Track & Small Claims at the ClaimIssued and ClaimDraft (which is missing from these changes) stages. Also, we definitely should not be removing it from TrialArrangementNotifyOther/Bundles etc Sql files, as those events will be run in Fast Track and the changes must be visible. Instead we could put extra checks in the handlers and run SQL scenarios that don't change trial arrangements.

miguelMolina3691 (approved):

miguelMolina3691 (dismissed):

kannan-v-hmcts (approved):

kannan-v-hmcts (approved):

miguelMolina3691 (approved):

dharmendrak (approved):

PR #4637, opened 2024-05-01
DTSCCI-000 fix npe in pip letter generator

ed14537:

https://tools.hmcts.net/jira/browse/DTSCCI-207

### Change description ###

Throw a dedicated exception. log the error message from the stitching service (with legacy case reference as that's all we have access to in place), remove some redundant tests that are testing null for values that ought to never be.

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/stitching/BundleRequestExecutor.java
```diff
@@ -55,10 +56,13 @@ public CaseData post(final BundleRequest payload, final String endpoint, String
...
             );
             if (response1.getStatusCode().equals(HttpStatus.OK)) {
-                return caseDetailsConverter.toCaseData(requireNonNull(response1.getBody()));
+                return Optional.of(caseDetailsConverter.toCaseData(requireNonNull(response1.getBody())));
             } else {
                 log.warn("The call to the endpoint with URL {} returned a non positive outcome (HTTP-{}). This may "
                              + "cause problems down the line.", endpoint, response1.getStatusCode().value());
+                log.info("Stitching endpoint returned {} with reason {}",
+                         response1.getStatusCodeValue(),
+                         response1.getStatusCode().getReasonPhrase());
```
rishikrsharma, line 65: if you add - case id that will be good.

srochani (dismissed):

src/main/java/uk/gov/hmcts/reform/civil/service/stitching/BundleRequestExecutor.java
ed14537: Where exactly do I get it from 🙈 - Nigel has already got his ticket to add it to all logs

srochani (dismissed):

src/main/java/uk/gov/hmcts/reform/civil/service/stitching/BundleRequestExecutor.java
rishikrsharma: From Payload -  if not then fine. :)

rishikrsharma (dismissed):

rishikrsharma (approved):

nigeldunne (approved):

PR #4685, opened 2024-05-09
CIV-13490 and civ-13491 intermediate DQs

drummondjm:
update responseClaimTrack allocation in RespondToDefenceSpecCallbackHandler

for https://tools.hmcts.net/jira/browse/CIV-13490 and
https://tools.hmcts.net/jira/browse/CIV-13491

```
[ ] Yes
[x ] No
```


LeonardoPalmeiro (approved):

sankaviv1 (approved):

Jenkinsfile_CNP
```diff
@@ -8,7 +8,7 @@ import uk.gov.hmcts.contino.GithubAPI
 def type = "java"
```
sherlynkhaw, line 8: you've added a new field in your ccd code called 
applicant1DQFixedRecoverableCostsIntermediate

Does it not need to be in caseData?

sherlynkhaw (changes requested):

Jenkinsfile_CNP
drummondjm: For these tickets are purely creating the CCD screens and config.

applicant1DQFixedRecoverableCostsIntermediate
respondent1DQFixedRecoverableCostsIntermediate
respondent2DQFixedRecoverableCostsIntermediate

Will need to be added to casedata in service, when we want to use them in the Docmosis outputs, but to me it made more sense to do that in those tickets when they are picked up, rather than these CCD tickets

drummondjm: I can update to add into casedata, will need done anyways, so either where it happens ticket wise.

Claim details wise, interesting point, no mention in tickets about that, makes sense though..

src/main/java/uk/gov/hmcts/reform/civil/model/dq/FixedRecoverableCosts.java
```diff
@@ -16,4 +17,7 @@ public class FixedRecoverableCosts {
     private ComplexityBand band;
     private YesOrNo complexityBandingAgreed;
     private String reasons;
+    // below used in intermediate claim FRC
+    private YesOrNo agreedWithOtherParty;
```
sherlynkhaw, line 21: Don't think this is needed. 
we can reuse this field `complexityBandingAgreed`

drummondjm: done

sherlynkhaw (approved):

LeonardoPalmeiro (approved):

sherlynkhaw (approved):

PR #4736, opened 2024-05-15
DTSCCI-206 user live epimmsid instead of closed epimmsid

rishikrsharma:
DTSCCI-206

### Change description ###
user live epimmsid instead of closed epimmsid

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/utils/LocationRefDataUtil.java
```diff
@@ -17,6 +19,18 @@
 public class LocationRefDataUtil {
 
     private final LocationRefDataService locationRefDataService;
+    /*
+    *  Below map holds the key = Closed court in ref data
+    *  and Value holds the live court against closed court
+    * */
+    public static final Map<String, String> liveEpimmsIdMap = Map.of("336348", "214320",
```
ruban72, line 26: Can we move this map to application.yaml rather than hardcoding?

velimd (approved):
code looks good to me but alot of numbers which im not sure the meaning of.

rishikrsharma:
> code looks good to me but alot of numbers which im not sure the meaning of.
in map key is courtId -court which has been closed, and value is replacement court.

PR #4820, opened 2024-05-30
DTSCCI-340 case note null check

rishikrsharma:
DTSCCI-340 case note null check


src/main/java/uk/gov/hmcts/reform/civil/service/robotics/mapper/EventHistoryMapper.java
```diff
@@ -821,16 +821,18 @@ private void buildMiscellaneousCaseNotesEvent(EventHistory.EventHistoryBuilder b
                          .dateReceived(caseNote.getCreatedOn())
                          .eventDetailsText(left((format(
                              "case note added: %s",
-                             caseNote.getNote().replaceAll("\\s+", " ")
+                             caseNote.getNote() != null
```
srochani, line 824: We can use apache common replaceAll which don't require null check

rishikrsharma: its existing code

ruban72 (approved):

PR #4890, opened 2024-06-10
CIV-13814 bilingual templates

dtortolaV1:
CIV-13814

### Change description ###
bilingual email templates

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -1480,6 +1488,32 @@ public boolean isPaidLessThanClaimAmount() {
...
+            && getApplicant1RepaymentOptionForDefendantSpec() == PaymentType.IMMEDIATELY;
+    }
+
+    /**
+     * LiP applicant requires Welsh if they selected Welsh or Both languages.
+     *
+     * @return true if applicant is LiP and they accept Welsh (or English+Welsh)
+     */
+    @JsonIgnore
+    public boolean isLipApplicantRequiringWelsh() {
```
m-meulendijks-v1, line 1500: Is this necessary? There should be a method isClaimantBilingual() and isRespondentResponseBilingual().
Don't they do the same thing?

It might look like they are not as they only look for Language. Both, but that's fine because only both or English is ever sent over from the CUI side (Both/Welsh/English is used for DQs only).

src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -1480,6 +1488,32 @@ public boolean isPaidLessThanClaimAmount() {
...
+        return Optional.ofNullable(getClaimantBilingualLanguagePreference())
+            .filter(value -> List.of(Language.WELSH.toString(), Language.BOTH.toString()).contains(value))
+            .isPresent();
+    }
+
+    /**
+     * LiP applicant requires Welsh if they selected Welsh or Both languages.
```
m-meulendijks-v1, line 1507: See comment above. If it does turn out we need this method, be sure that the comment reflects the use of respondent rather than Applicant.

m-meulendijks-v1 (dismissed):

laarmada (dismissed):

dharmendrak (approved):

PR #4998, opened 2024-06-27
DTSCCI-402 - 

ed14537:

https://tools.hmcts.net/jira/browse/DTSCCI-402

### Change description ###

Global exception handler 
Add caching of request data as thread decorator to easily access case_id and auth tokens in classes without changing method signatures
Add caseIds to error logs in current exception handlers using above

```
[ ] Yes
[x] No
```


velimd (dismissed):
LGTM, I dont know if you need to format the error messages like that as I thought you can do it with logger itself but its fine either way.

rishikrsharma (approved):

Jefex7 (dismissed):

Jefex7 (approved):

PR #5216, opened 2024-08-01
CIV-14740 carm date change

drummondjm:
https://tools.hmcts.net/jira/browse/CIV-14740

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowPredicate.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowPredicate.java
@@ -423,7 +423,7 @@ public class FlowPredicate {
 
     private static boolean getCarmEnabledForDate(CaseData caseData) {
         // Date of go live is 1st August, as we use "isAfter" we compare with 31st July
-        return caseData.getSubmittedDate().toLocalDate().isAfter(LocalDate.of(2024, 7, 31));
+        return caseData.getSubmittedDate().toLocalDate().isAfter(LocalDate.of(2024, 11, 4));
     }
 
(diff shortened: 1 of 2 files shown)


sherlynkhaw (dismissed):

rishikrsharma (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/flowstate/FlowPredicate.java
```diff
@@ -422,8 +422,8 @@ private static boolean isSpecSmallClaim(CaseData caseData) {
     }
 
     private static boolean getCarmEnabledForDate(CaseData caseData) {
-        // Date of go live is 1st August, as we use "isAfter" we compare with 31st July
-        return caseData.getSubmittedDate().toLocalDate().isAfter(LocalDate.of(2024, 7, 31));
+        // Date of go live is  5th november , as we use "isAfter" we compare with 4th november
+        return caseData.getSubmittedDate().toLocalDate().isAfter(LocalDate.of(2024, 8, 8));
```
dtortolaV1, line 426: Shouldn't we use a constant reachable by both classes, so we only have to check one?
Also, StateFlowEngine:635 transitions to mediation when a case is spec && small claim && carmEnabledForDate, even if one of the parties specifically said they don't want mediation

drummondjm: don't need a constant for now as the temporary duplicate classes will be cleared by ci squad as part of their clean up activity . .Ambigious state will be cleaned up later on by ci squad "Longer term, will need to solve the ambiguous state transition"


@

PR #5463, opened 2024-09-19
CIV-15222 Update View Bundle and Trial Arrangements task List when Bundle is Restitched

kannan-v-hmcts:
PR for https://tools.hmcts.net/jira/browse/CIV-15222 & https://tools.hmcts.net/jira/browse/CIV-15382


m-meulendijks-v1 (approved):

src/main/resources/db/migration/V2024_09_20_1100__CIV-15222_bundle_restitch_task_list_scenario.sql
```diff
@@ -0,0 +1,38 @@
+/**
```
ss-evoco, line 1: If these need running separately from the rest in the folder (for a separate release) then I think they should be in a new folder now.

src/main/resources/db/migration/V2024_09_23_1200__CIV-15382_bundle_restitch_remove_full_stop.sql
```diff
@@ -0,0 +1,10 @@
+
```
ss-evoco, line 1: Same with this.

ss-evoco (changes requested):

src/main/resources/db/migration/V2024_09_20_1100__CIV-15222_bundle_restitch_task_list_scenario.sql
kannan-v-hmcts: HI Sean. Thanks for your review. As far as I know and I could be wrong as well, I could not find any config that tell the flyway DB to look for scripts in a specific folder so I am assuming it takes it from the default folder ie classpath:db/migration and the migration is enabled/disabled using REFERENCE_DATABASE_MIGRATION. At the moment its enabled in preview but not in demo/prod. its enabled once when a release happens and disabled again.

kannan-v-hmcts: @dharmendrak Could you please confirm if this is right ?

ss-evoco (approved):

src/main/resources/db/migration/V2024_09_23_1200__CIV-15382_bundle_restitch_remove_full_stop.sql
ss-evoco: Agreed to do this in a later, separate ticket.

PR #5580, opened 2024-10-09
CIV-15087 - COSC Application Processor Scheduler

Gareth40343:
- Added process cosc application logic, events and handlers

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/civil/service/search/CoscApplicationSearchService.java
```diff
@@ -0,0 +1,37 @@
...
+public class CoscApplicationSearchService extends ElasticSearchService {
+
+    public CoscApplicationSearchService(CoreCaseDataService coreCaseDataService) {
+        super(coreCaseDataService);
+    }
+
+    public Query query(int startIndex) {
+        return new Query(
+            boolQuery()
+                .must(matchQuery("data.coSCApplicationStatus", "Active"))
```
Gareth40343, line 24: TODO: Confirm query parameters!!!

Azam-Hmcts: I think we also need to consider coscShedulerDeadline. .coSCApplicationPaidInFull And state is not required. But we can discuss about state

sherlynkhaw (approved):

ShwetaTandel-hmcts (approved):

PR #5732, opened 2024-10-28
CIV-15300 Solve bug task list update

miguelMolina3691:
https://tools.hmcts.net/jira/browse/CIV-15300

### Change description ###
Before case proceeds on caseman:

CIV-15300-before

After case proceeds on caseman:

CIV-15300-after

```
[ ] Yes
[ x ] No
```


src/main/resources/db/migration/V2024_10_25_1230__CIV-15300_case_proceed_caseman_task_list_update.sql
```diff
@@ -0,0 +1,29 @@
+/**
```
ss-evoco, line 1: Will this have any knock on effects to accounts in production in this state ? Will all their task states just disappear for example ?

ss-evoco (changes requested):

PR #5769, opened 2024-11-05
DTSCCI-1093 fix for debit data null check

rishikrsharma:
DTSCCI-1093

### Change description ###
fix for debit data null check

```
[ ] Yes
[ ] No
```


gmikerlemos (approved):
What do you think about using Optional.ofNullable(debtLRspec.getPaymentAmount()).orElse(BigDecimal(0)), or maybe extract this or even create a Utility method like zeroIfNull().

ruban72 (approved):

PR #5788, opened 2024-11-08
DTSCCI-1137 Protect against NPE

nigeldunne:
https://tools.hmcts.net/jira/browse/DTSCCI-1137

### Change description ###

DTSCCI-1137: Protect against NPE

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/civil/model/docmosis/common/ReasonMoneyTemplateData.java
```diff
@@ -94,7 +94,12 @@ public static ReasonMoneyTemplateData toReasonMoneyTemplateData(RecurringExpense
             .type(item.getType() == ExpenseTypeLRspec.OTHER
                       ? "Other: " + item.getTypeOtherDetails()
                       : EXPENSE_TYPE_LIP_RESPONSE.get(item.getType()))
-            .amountPounds((MonetaryConversions.penniesToPounds(item.getAmount())).setScale(2, RoundingMode.CEILING))
+            .amountPounds(getAmountPounds(item.getAmount()))
             .build();
     }
+
+    private static BigDecimal getAmountPounds(final BigDecimal item) {
+        return (item == null ? null : MonetaryConversions.penniesToPounds(item).setScale(2, RoundingMode.CEILING));
```
rishikrsharma, line 102: if null then i would say set new BigDecimal(0) rather than null

velimd (approved):

src/main/java/uk/gov/hmcts/reform/civil/model/docmosis/common/ReasonMoneyTemplateData.java
rishikrsharma: this is just utility class  and i would prefer 0 rather than null. I deally it should call this method itself if value is nulll

nigeldunne: That would make it look like the user had entered a value but they haven't

rishikrsharma (approved):

velimd (approved):

PR #5839, opened 2024-11-23
DTSCCI-1202 adding missing submitted callback 

rishikrsharma:
DTSCCI-1202

### Change description ###
Adding missing submitted callback as this is required for EventEmitterAspect which looks for Submitted event

```
[ ] Yes
[ ] No
```


ed14537 (approved):

ed14537 (commented):
Do all of these events trigger business processes? I think it's only handlers that invoke processes (set them to ready) that require the empty submitted callbacks

PR #5874, opened 2024-12-04
DTSCCI-1273 lip response form null check

rishikrsharma:
DTSCCI-1273

### Change description ###
lip response form null check

```
[ ] Yes
[ ] No
```


Jefex7 (approved):

velimd (commented):
maybe we should have updated unit test for these changes?

PR #6640, opened 2025-06-12
CIV-17247 RJBA - Fixed cost + judgment amount

kalachandrasekar1:

https://tools.hmcts.net/jira/browse/CIV-17247

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/notification/handlers/CamundaProcessIdentifier.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/notification/handlers/CamundaProcessIdentifier.java
@@ -21,6 +21,7 @@ public enum CamundaProcessIdentifier {
     JudgmentVariedDeterminationOfMeansNotifyParties,
     ClaimSubmittedApplicantNotifier,
     ClaimantLipHelpWithFeesNotifier,
+    TakeCaseOfflineNotifier,
     UnpaidHearingFeeNotifier,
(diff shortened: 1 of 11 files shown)


m-meulendijks-v1 (dismissed):

miguelMolina3691 (approved):

miguelMolina3691 (dismissed):

m-meulendijks-v1 (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/JudgementService.java
```diff
@@ -160,11 +160,6 @@ && isLRvLR(caseData)
     }
 
     private YesOrNo checkFixedCostOption(CaseData caseData) {
-        if (isLrFullAdmitRepaymentPlan(caseData)
```
m-meulendijks-v1, line 163: I think Miguel added this recently as part of another Fixed Costs ticket, can we compare tickets so that we know we're not accidentally removing something relevant?

m-meulendijks-v1 (approved):

src/main/java/uk/gov/hmcts/reform/civil/service/JudgementService.java
kalachandrasekar1: If the fixed cost option is selected during claim creation and not selected in judgment flow then ccjJudgmentFixedCostOption should be 'No'. That's the reason it's removed

PR #6935, opened 2025-08-20
DTSCCI-2627 Blocking Bug Fix

gergelykisshmcts:

https://tools.hmcts.net/jira/browse/DTSCCI-2627

### Change description ###

The Problem (Root Cause Analysis)

PR introducing the blocking testing bug: #6772
https://github.com/hmcts/civil-service/pull/6772

The test was failing because it expected a hardcoded future date:
Expected: "queryDate" = "20 August 2025"
Actual:   "queryDate" = "19 August 2025"

The issue was in the test code using OffsetDateTime.now() which returns
the current system time:
- When the test runs on August 19, it returns "19 August 2025"
- When the test runs on August 20, it would return "20 August 2025"
- The test was written expecting it to always be "20 August 2025"

Found an issue in QueryResponseSolicitorNotificationHandlerTest.java
where multiple test methods used:
OffsetDateTime.now() <-This changes based on when the test runs...

Replaced all instances of OffsetDateTime.now() with a fixed timestamp:
// Before (please never do this again):
OffsetDateTime.now()

// After (stable):
OffsetDateTime.of(2025, 8, 20, 10, 0, 0, 0, ZoneOffset.UTC)

Affected tests: All test methods that were expecting "20 August 2025"
in their assertions

Best Practice:
Standard testing principle: Never use current time in tests. Always use:
- Fixed timestamps for predictable results
- Mock time services when testing time-dependent logic
- Test utilities that control time flow
- What can we learn from this: Never write during testing or approve during code review tests thats sometimes passes and sometimes fails based on external factors (in this
  case, the system clock).

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/QueryResponseSolicitorNotificationHandlerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/civil/handler/callback/camunda/notification/QueryResponseSolicitorNotificationHandlerTest.java
@@ -36,6 +36,7 @@ import uk.gov.hmcts.reform.civil.service.querymanagement.QueryManagementVariable
 
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


gmikerlemos (changes requested):
I think this is not needed, I have just run a clean build and the test is passing as it is.

PR #7997, opened 2026-06-17
DTSCCI-133 - 1v2 different solicitors defendant response forms incorr…

gajanan-deshpande:
…… (#7739)

https://tools.hmcts.net/jira/browse/DTSCCI-133
https://tools.hmcts.net/jira/browse/DTSCCI-5671

### Change description ###
DTSCCI-133 - 1v2 different solicitors defendant response forms incorrectly generated and fix Lip vs Lip journey

```
[ ] Yes
[ ] No
```


Jenkinsfile_CNP
```diff
@@ -340,7 +340,7 @@ def setPreviewUrls(String product, String component) {
     env.PREVIEW_HMC_API_URL = wiremockUrl
     env.PREVIEW_RD_PROFESSIONAL_API_URL = wiremockUrl
     env.PREVIEW_RD_COMMONDATA_API_URL = "http://rd-commondata-api-aat.service.core-compute-aat.internal"
-    env.PREVIEW_DOCMOSIS_TORNADO_URL = wiremockUrl
```
sbagga-moj, line 343: do we need this change?

Jenkinsfile_CNP
```diff
@@ -340,7 +340,7 @@ def setPreviewUrls(String product, String component) {
...
+    env.PREVIEW_DOCMOSIS_TORNADO_URL = "https://docmosis.aat.platform.hmcts.net"
```
gmikerlemos, line 343: Is this Docmosis preview URL change intentional? This makes preview functional tests call the shared AAT Docmosis endpoint instead of wiremock, unlike the other services in this block, which could make PR validation depend on AAT availability/configuration.

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/spec/response/confirmation/RepayPlanConfirmationText.java
```diff
@@ -26,8 +26,12 @@ public Optional<String> generateTextFor(CaseData caseData, FeatureToggleService
 
         if (!RespondentResponsePartAdmissionPaymentTimeLRspec.SUGGESTION_OF_REPAYMENT_PLAN.equals(
             caseData.getDefenceAdmitPartPaymentTimeRouteRequired())
+            && !RespondentResponsePartAdmissionPaymentTimeLRspec.SUGGESTION_OF_REPAYMENT_PLAN.equals(
+            caseData.getDefenceAdmitPartPaymentTimeRouteRequired2())
```
gmikerlemos: This looks like it can match stale data from the other respondent in 1v2 different-solicitor cases. If R1 previously selected a repayment plan and R2 is now submitting a different response/payment route, this OR check will still make R2 hit the repayment-plan confirmation. Can we scope this to the currently submitting respondent, using `isRespondent1`/`isRespondent2` or a current-respondent accessor, instead of checking either respondent field globally?

src/main/java/uk/gov/hmcts/reform/civil/service/docmosis/sealedclaim/SealedClaimResponseFormGeneratorForSpec.java
```diff
@@ -166,13 +164,12 @@ private void handleDefenceResponseDocument(SealedClaimResponseFormForSpec form,
...
 
     private void handlePayments(CaseData caseData, SealedClaimResponseFormForSpec form) {
-        Stream.of(caseData.getRespondToClaim(), caseData.getRespondToAdmittedClaim())
-            .filter(Objects::nonNull)
-            .findFirst()
-            .ifPresent(response -> form
-                .setPoundsPaid(MonetaryConversions.penniesToPounds(response.getHowMuchWasPaid()).toString())
-                .setPaymentDate(response.getWhenWasThisAmountPaid())
-                .setPaymentMethod(getPaymentMethod(response)));
+        RespondToClaim respondToClaim = caseData.getResponseToClaim();
```
gmikerlemos: `getResponseToClaim()` now returns a response only when `isRespondent1` or `isRespondent2` is explicitly set. The separate `GENERATE_RESPONSE_SEALED` callback appears to identify DEF2 via `respondent2DocumentGeneration`, not by setting these flags, so payment fields can be omitted if the flags are absent or stale. Could this selection be based on the same respondent marker used by the sealed-form generation path, or retain a safe 1v1/default fallback?

Jenkinsfile_CNP
gajanan-deshpande: This is needed for testing as PDF will be checked by QA to see if correct values are shown after generation. I will remove it before merging to master

gmikerlemos: The QAs test usually are done in a CUI pr with fullDeployment label. So I wouldn't change this here and risk this being forgotten

src/main/java/uk/gov/hmcts/reform/civil/model/docmosis/sealedclaim/ResponseRepaymentDetailsForm.java
```diff
@@ -157,8 +156,20 @@ private static void addPayBySetDate(CaseData caseData, ResponseRepaymentDetailsF
     }
 
     private static void addPayByDatePayImmediately(ResponseRepaymentDetailsFormData data, BigDecimal totalClaimAmount, CaseData caseData) {
-        LocalDate whenWillThisAmountBePaid = Optional.ofNullable(caseData.getRespondToClaimAdmitPartLRspec()).map(
-            RespondToClaimAdmitPartLRspec::getWhenWillThisAmountBePaid).orElse(null);
+        LocalDate whenWillThisAmountBePaid = null;
+        if (YES.equals(caseData.getIsRespondent1()) && caseData.getRespondToClaimAdmitPartLRspec() != null) {
```
gmikerlemos: This is another instance of respondent-selection logic being implemented locally. It would be easier to reason about if the method first selected the relevant respondent data once, then mapped that to `whenWillThisAmountBePaid`. The current pattern mixes null checks, respondent flags and fallback behaviour in the document mapper.

src/main/java/uk/gov/hmcts/reform/civil/model/CaseData.java
```diff
@@ -744,29 +744,40 @@ public boolean isPayBySetDate() {
 
     @JsonIgnore
     public boolean isPayByInstallment() {
-        return defenceAdmitPartPaymentTimeRouteRequired != null
+        return (defenceAdmitPartPaymentTimeRouteRequired != null
             && defenceAdmitPartPaymentTimeRouteRequired
-            == RespondentResponsePartAdmissionPaymentTimeLRspec.SUGGESTION_OF_REPAYMENT_PLAN;
+            == RespondentResponsePartAdmissionPaymentTimeLRspec.SUGGESTION_OF_REPAYMENT_PLAN)
+            || (getDefenceAdmitPartPaymentTimeRouteRequired2() != null
```
gmikerlemos, line 750: This helper now has a broader meaning than the method name suggests: it returns true if either respondent selected instalments. In 1v2 different-solicitor flows that makes callers depend on global case state rather than the current respondent. For readability and safety, can we either keep this as the original respondent-1/default check and add a respondent-scoped helper, or make the current respondent explicit in the method name/signature?

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/spec/response/confirmation/PartialAdmitPayImmediatelyConfirmationText.java
```diff
@@ -30,13 +31,21 @@ public class PartialAdmitPayImmediatelyConfirmationText implements RespondToClai
...
+            caseData.getDefenceAdmitPartPaymentTimeRouteRequired())
+            && !RespondentResponsePartAdmissionPaymentTimeLRspec.IMMEDIATELY.equals(
+            caseData.getDefenceAdmitPartPaymentTimeRouteRequired2())) {
             return Optional.empty();
         }
         LocalDate whenBePaid = Optional.ofNullable(caseData.getRespondToClaimAdmitPartLRspec())
             .map(RespondToClaimAdmitPartLRspec::getWhenWillThisAmountBePaid)
             .orElse(null);
 
+        if (YES.equals(caseData.getIsRespondent2())) {
```
gmikerlemos: This defaults to respondent 1 and then conditionally overrides with respondent 2. That makes the current respondent selection implicit and easy to get wrong as more 1v2 scenarios are added. Could this be centralised behind a helper such as `getCurrentRespondentAdmitPartLrSpec()` / `getCurrentRespondentPaymentDate()` so the confirmation text does not need to know which respondent field to read?

src/main/java/uk/gov/hmcts/reform/civil/model/docmosis/sealedclaim/ResponseRepaymentDetailsForm.java
```diff
@@ -218,7 +236,10 @@ private static void fullDefenceData(CaseData caseData, ResponseRepaymentDetailsF
 
     private static void partAdmissionData(CaseData caseData, ResponseRepaymentDetailsFormData data) {
         addDetailsOnWhyClaimIsRejected(caseData, data);
-        if (caseData.getSpecDefenceAdmittedRequired() == YesOrNo.YES) {
+        if (YES.equals(caseData.getIsRespondent1()) && caseData.getSpecDefenceAdmittedRequired() == YesOrNo.YES
```
gmikerlemos: This condition is hard to review safely because it mixes respondent identity, admitted flags, response dates, `&&` and `||` in one expression. Could we split this into named booleans, e.g. `isCurrentRespondentAlreadyPaid` and `isLegacyRespondent1AlreadyPaidFallback`, so the intended cases are explicit?

src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/spec/response/confirmation/PartialAdmitPaidFullConfirmationText.java
```diff
@@ -21,11 +20,12 @@ public class PartialAdmitPaidFullConfirmationText implements RespondToClaimConfi
...
         if (!RespondentResponseTypeSpec.PART_ADMISSION.equals(caseData.getRespondent1ClaimResponseTypeForSpec())
-            || NO.equals(caseData.getSpecDefenceAdmittedRequired())) {
+            && !RespondentResponseTypeSpec.PART_ADMISSION.equals(caseData.getRespondent2ClaimResponseTypeForSpec())
+            || NO.equals(caseData.getSpecDefenceAdmittedRequired())
+            && NO.equals(caseData.getSpecDefenceAdmitted2Required())) {
             return Optional.empty();
         }
-        BigDecimal howMuchWasPaid = Optional.ofNullable(caseData.getRespondToAdmittedClaim())
-            .map(RespondToClaim::getHowMuchWasPaid).orElse(null);
+        BigDecimal howMuchWasPaid = (caseData.getResponseToClaim() != null ? caseData.getResponseToClaim().getHowMuchWasPaid() : null);
```
gmikerlemos: Minor maintainability issue: `getResponseToClaim()` is nullable and is called twice here. More importantly, the caller has to know that this helper chooses the current respondent. Consider assigning it once, or exposing a more explicit current-respondent paid amount helper so the null handling and respondent selection are not repeated in each confirmation-text generator.

PR #8047, opened 2026-06-25
DTSCCI-5726: Harden document-download error handling (differentiate CDAM failures)

ruban72:
### JIRA link ###
https://tools.hmcts.net/jira/browse/DTSCCI-5726 (also addresses DTSCCI-5727 and EXC-CS-008/029/047)

### Change description ###

`SecuredDocumentManagementService`'s three download methods each ended in a blanket
`catch (Exception ex) -> throw new DocumentDownloadException(documentPath, ex)`, which
collapsed every distinct failure into one opaque exception that fell through to a generic
**HTTP 500**. Two failure modes were also amplified by `@Retryable(maxAttempts = 5)` retrying
deterministic failures that can never succeed.

This PR differentiates the failure modes so deterministic failures return a meaningful status
without retrying, while genuinely transient failures still retry.

**Changes**
- New exceptions: `DocumentNotFoundException` (404), `DocumentAccessException` (403),
  `InvalidDocumentReferenceException` (400), mapped in `ResourceExceptionHandler` (mirrors the
  existing typed-Feign handlers).
- Download methods now catch `FeignException.NotFound` / `FeignException.Forbidden` /
  `IllegalArgumentException` distinctly; transient 5xx/504 stays the retryable
  `DocumentDownloadException`. Deterministic types are excluded via `@Retryable(noRetryFor = ...)`,
  and already-typed exceptions are rethrown to avoid re-wrapping.
- `DocumentDownloadService` guards null/blank document id before it reaches CDAM or the retry
  loop (stops the `documents/null` 5x-retry storm).
- Unit tests for 404/403 differentiation and the null/blank id guard.

**Production issues addressed**

| Exception | Before | After |
|---|---|---|
| EXC-CS-010 (DTSCCI-5726) wrapper | known 404 -> 500 | 404 -> 404 |
| EXC-CS-009 (DTSCCI-5727) CDAM 404 | 500, retried 5x | 404, no retry |
| EXC-CS-029 null document id | 500, retried 5x | 400, no CDAM call |
| EXC-CS-047 CDAM 403 TTL | 500, retried | 403, no retry storm |
| EXC-CS-008 CDAM 504 | 500 | stays retryable transient |

**Out of scope:** EXC-CS-078 (OOM streaming, library-level) and the underlying data cleanup
for EXC-CS-009 / EXC-CS-047 (orphaned refs / TTL) are separate follow-ups.

**Does this PR introduce a breaking change?**

```
[ ] Yes
[x] No
```

Note for reviewers: `DocumentDownloadException` (genuinely-unknown / transient) still maps to 500
via `UncaughtExceptionHandler` and stays retryable. Flagging in case you'd prefer it mapped to
502 Bad Gateway for upstream CDAM failures.


src/main/java/uk/gov/hmcts/reform/civil/documentmanagement/SecuredDocumentManagementService.java
```diff
@@ -158,6 +159,8 @@ public CaseDocument uploadDocument(String authorisation, UploadedDocument upload
     }
 
     @Retryable(retryFor = {DocumentDownloadException.class},
+        noRetryFor = {DocumentNotFoundException.class, DocumentAccessException.class,
+            InvalidDocumentReferenceException.class},
         maxAttempts = 5,
```
phil-milne: Not this change but related:  There is significant failed retry traffic in the logs, can we reduce the 5 attempts to 3 and increase the multiplier from 2 to 3, for upload and download now that the exceptions have been refined? This would help with churn from FeignException.GatewayTimeout | FeignException.BadGateway | FeignException.ServiceUnavailable etc..

ruban72: Good shout, done in 03aaa5dea2 - maxAttempts 5->3 and backoff multiplier 2->3 on both upload and download `@Retryable` methods. Bad-input cases (404/403/malformed) are already excluded via `noRetryFor`/classification, so this just trims the genuine transient-failure backoff (GatewayTimeout/BadGateway/ServiceUnavailable).

phil-milne (approved):
LGTM

PR #8101, opened 2026-07-08
DTSCCI-5823: add S2S secret readiness health indicator

ruban72:
## What
Adds `S2sSecretReadinessHealthIndicator` — a diagnostic health check that verifies **civil-service** can actually *use* its mounted secret material, not just that a secret file is present.

It mints an S2S token via the existing `serviceAuthTokenGenerator` (backed by `idam.s2s-auth.totp_secret` from the Key Vault CSI mount). If that secret failed to mount or is stale, token generation fails and the indicator reports `DOWN` on its own `/health/s2s` group for alerting.

## Why
This is the failure mode behind the 30 Jun P1 (incident 179051 / DTSCCI-5762): a single replica came up with stale secret material and rejected ~1 in 9 CCD callbacks with 403 for ~19h while still reporting Ready. This is the Civil CI service-layer belt-and-braces measure from DTSCCI-5823 (MIR 5); the durable platform fix for the Key Vault / DNS / CSI mount failure remains with Platform Operations (MIR 1/3).

## Design note: diagnostic signal, not a readiness gate
An earlier revision gated the Kubernetes readiness probe on this check. That was **deliberately changed**. Coupling readiness to a live external S2S call means a transient S2S blip pulls *every* pod out of rotation at once — a fleet-wide outage, which is worse than the intermittent 403s it aims to catch. So the check is exposed on its own `/health/s2s` group for alerting/dashboards and is **not** in the `readiness` group. If automatic remediation for a genuinely stale secret is wanted, the recommended path is an alert on `/health/s2s` DOWN (paging / targeted restart), not the readiness probe.

## How
- The S2S call runs on a `@Scheduled` fixed-delay (default 120s) and the outcome is cached, so any health poll only reads in-memory state and never hammers the S2S service. Periodic (not start-up only) to catch a pod that goes stale after a healthy start via the ~9-minute CSI rotation.
- **Optimistic start:** reports `UP` until proven `DOWN`, so a freshly deployed pod is not flagged during the start-up window before its first check.
- **Blip tolerance:** flips to `DOWN` only after `civil.health.s2s.failure-threshold` (default 2) consecutive failures; recovers to `UP` on the next success.
- **Always registered:** the bean is always present (no `@ConditionalOnProperty`), so the `s2s` health-group membership stays valid in every configuration. When `civil.health.s2s.enabled=false` the check reports `UP` and skips the scheduled call rather than being removed (which would fail startup under Spring Boot group-membership validation).
- Exposed via the management health group `s2s` (`/health/s2s`); not wired to liveness or readiness.
- Config knobs (all with in-code defaults): `civil.health.s2s.enabled` / `check-interval-ms` / `initial-delay-ms` / `failure-threshold`.

## Tests
- `S2sSecretReadinessHealthIndicatorTest` — up-before-first-check (optimistic), up-after-success, stays-up-on-single-failure, down-after-threshold, down-after-threshold-blank-tokens, blip-tolerance-then-down, recovery-to-up, and disabled-reports-up-never-calls-s2s.
- `S2sSecretReadinessHealthGroupIntegrationTest` — the `s2s` group includes the contributor and goes UP/DOWN with it, and the `readiness` group does **not** include it (proves no traffic gating).
- `S2sSecretReadinessDisabledStartupIntegrationTest` — the context starts with the check disabled while the `s2s` group still references it under group-membership validation (regression guard for the disabled config).

All passing (`./gradlew test`).


mapingo (approved):

src/test/java/uk/gov/hmcts/reform/civil/health/S2sSecretReadinessHealthGroupIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+        markContextAcceptingTraffic();
+        when(serviceAuthTokenGenerator.generate())
+            .thenThrow(new IllegalStateException("microservice key is null"));
+        indicator.refresh();
+
+        HealthComponent readiness = healthEndpoint.healthForPath("readiness");
+
+        assertThat(((CompositeHealth) readiness).getComponents())
+            .as("the S2S check must NOT gate the Kubernetes readiness probe")
+            .doesNotContainKey("s2sSecretReadiness");
```
gmikerlemos, line 129: Could we align this with the intended outcome before merging ? The PR description says this check will remove a pod from readiness but the implementation and this test explicitly keep readiness up when S2S is down  .That means a pod with a stale secret will continue receiving CCD callbacks which is the incident this change is meant to prevent , If the decision is now for this to be diagnostic only I think the title description and monitoring need updating and an alert consuming the S2S health group should be part of the delivery

src/main/resources/application.yaml
```diff
@@ -7,6 +7,9 @@ management:
       show-details: "always"
       probes:
         enabled: true
+      group:
+        s2s:
+          include: s2sSecretReadiness
```
gmikerlemos, line 12: Shouldn't we cover the disabled configuration here,  the bean is removed when civil health s2s enabled is false but this group still includes s2sSecretReadiness ,Spring Boot validates health group membership by default so the application will fail to start because the contributor does not exist.  We should either keep the contributor registered when disabled remove the enabled option or make the group configuration conditional and add a disabled context test

ruban72: Good catch, thanks. Fixed in b8f74aa: removed the `@ConditionalOnProperty` so the indicator bean is always registered, which keeps the `s2s` group membership valid in every configuration. When `civil.health.s2s.enabled=false` the check now reports UP and skips the scheduled S2S call instead of being removed.

Added a startup regression test (`S2sSecretReadinessDisabledStartupIntegrationTest`) that boots the context with `enabled=false` + `validate-group-membership=true` so this can't regress.

src/test/java/uk/gov/hmcts/reform/civil/health/S2sSecretReadinessHealthGroupIntegrationTest.java
ruban72: You're right that the description is stale, I'll update it. This was a deliberate change rather than an oversight.

Gating the Kubernetes readiness probe on a live S2S call means a transient S2S blip pulls every pod out of rotation at once (a fleet-wide outage), which is worse than the 403s it was trying to avoid. So the indicator is now a diagnostic signal on its own `/health/s2s` group for alerting, and deliberately not in the `readiness` group; the test asserting readiness stays UP when S2S is down is intentional.

If we do want automatic remediation for a genuinely stale secret, I'd suggest an alert on `/health/s2s` DOWN (paging or a targeted restart) rather than the readiness probe. Happy to switch the gating behaviour back if the team would prefer it, but I'd argue against coupling readiness to an external dependency.

src/main/java/uk/gov/hmcts/reform/civil/health/S2sSecretReadinessHealthIndicator.java
```diff
@@ -0,0 +1,145 @@
...
+ * ~9 minute rotation, so this check is periodic (not start-up only) to catch a pod that goes
+ * stale after a healthy start.</p>
+ *
+ * <p>The bean is always registered (so the {@code s2s} health group membership stays valid and the
+ * application starts in every configuration). When {@code civil.health.s2s.enabled} is false the
+ * check is inert: it reports UP and skips the scheduled S2S call, rather than removing the bean.</p>
+ */
+@Slf4j
+@Component
+public class S2sSecretReadinessHealthIndicator implements HealthIndicator {
```
gmikerlemos, line 43: As a suggestion we could we rename this to S2sSecretHealthIndicator or S2sSecretDiagnosticHealthIndicator. It deliberately does not contribute to readiness now, so keeping Readiness in the class name may be confusing later.

gmikerlemos (approved):
lgtm only one non-blocking naming suggestion.

PR #8131, opened 2026-07-16
DTSCCI-5818 Docmosis document-generation null-guard failures

matt-goodwin-solirius:

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

https://tools.hmcts.net/jira/browse/DTSCCI-5818

### Change description ###

Docmosis document-generation null-guard failures

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/GenerateDirectionOrderCallbackHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/GenerateDirectionOrderCallbackHandler.java
@@ -273,7 +273,7 @@ public class GenerateDirectionOrderCallbackHandler extends CallbackHandler {
     private CallbackResponse generateTemplate(CallbackParams callbackParams) {
         CaseData caseData = callbackParams.getCaseData();
 
-        if (!BLANK_TEMPLATE_TO_BE_USED_AFTER_A_HEARING.equals(caseData.getFinalOrderDownloadTemplateOptions().getValue().getLabel())) {
+        if (!BLANK_TEMPLATE_TO_BE_USED_AFTER_A_HEARING.equals(getFinalOrderDownloadTemplateOptionLabel(caseData))) {
             CaseDocument documentDownload = judgeOrderDownloadGenerator.generate(caseData, callbackParams.getParams().get(BEARER_TOKEN).toString());
             caseData.setFinalOrderDownloadTemplateDocument(documentDownload);
             caseData.setShowOrderAfterHearingDatePage(NO);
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/reform/civil/handler/callback/user/GenerateDirectionOrderCallbackHandler.java
```diff
@@ -273,7 +273,7 @@ private CallbackResponse populateAndResetPreviousSelections(CaseData caseData) {
     private CallbackResponse generateTemplate(CallbackParams callbackParams) {
         CaseData caseData = callbackParams.getCaseData();
 
-        if (!BLANK_TEMPLATE_TO_BE_USED_AFTER_A_HEARING.equals(caseData.getFinalOrderDownloadTemplateOptions().getValue().getLabel())) {
+        if (!BLANK_TEMPLATE_TO_BE_USED_AFTER_A_HEARING.equals(getFinalOrderDownloadTemplateOptionLabel(caseData))) {
             CaseDocument documentDownload = judgeOrderDownloadGenerator.generate(caseData, callbackParams.getParams().get(BEARER_TOKEN).toString());
```
gmikerlemos, line 277: This still looks like it will fail when finalOrderDownloadTemplateOptions is missing. The null safe helper gets us past this check, but we then call judgeOrderDownloadGenerator.generate and JudgeOrderDownloadGenerator.getDownloadTemplate dereferences the same missing field. The new test mocks the generator, so it does not cover the production path. Could we handle the missing option inside the generator or return a controlled error here, and add a test that exercises the real generator path?

src/main/java/uk/gov/hmcts/reform/civil/utils/DocmosisTemplateDataUtils.java
```diff
@@ -158,29 +159,35 @@ public static List<String> fetchResponseIntentionsDocmosisTemplate(CaseData case
...
+                responseIntentions.add(labelOrEmpty(caseData.getRespondent1ClaimResponseIntentionType()));
                 return responseIntentions;
         }
         return responseIntentions;
     }
 
+    private static String labelOrEmpty(ResponseIntention responseIntention) {
+        return ofNullable(responseIntention)
+            .map(ResponseIntention::getLabel)
+            .orElse("");
```
gmikerlemos, line 188: Could we confirm with the business that a blank response intention is acceptable in the acknowledgement document? This avoids the exception, but it can produce text such as Defendant 1 with no response shown. If that is the agreed fallback, I am happy with this approach. Otherwise, a clear fallback value may be safer.

gmikerlemos (approved):
Re-reviewed after the latest changes. The downstream null path is now handled in JudgeOrderDownloadGenerator and covered by a test using the real generator path. No blocking issues remain. I left one non-blocking question about the agreed wording when a response intention is missing.

src/main/java/uk/gov/hmcts/reform/civil/utils/DocmosisTemplateDataUtils.java
matt-goodwin-solirius: Resolved with Rob, can be left blank

PR #8135, opened 2026-07-16
DTSCCI-5548 Persist draft claims in the CMC database for 180-day recovery

AbuSalamHMCTS:
https://tools.hmcts.net/jira/browse/DTSCCI-5548

## Summary

This work adds durable database storage for Citizen UI draft claims in the CMC PostgreSQL database. Draft claims can now be created, retrieved, updated and deleted through authenticated `civil-service` APIs without using Redis as the source of truth.

The draft-store implementation is reusable for future draft types. Generic persistence is handled by the draft-store service, while claim-specific behaviour remains in the draft-claim service. Each draft type can define its own retention period, with draft claims currently expiring 180 days after they are first created.

The original `draftClaimCreatedAt` and `expiresAt` values are kept when a draft is updated, so saving a draft does not extend its expiry date. Expired drafts cannot be retrieved or updated.

The implementation maintains the existing Citizen UI behaviour of one active claim draft per user. Repeated creation returns the existing active draft without changing it, while an expired draft is removed and replaced.

The new API provides the database storage needed for `civil-citizen-ui` to save and recover claim issue drafts when Redis data is unavailable. It also provides an owner-scoped deletion API for `civil-citizen-ui` to call after a claim is successfully submitted.

## Changes

- Added a Flyway migration to create the `draft_type` and `draft_store` tables.
- Added generic storage for draft ID, user ID, optional case ID, draft type, JSON payload, draft creation time, expiry time and audit timestamps.
- Used the generic `draft_created_at` database column while keeping `draftClaimCreatedAt` in the draft-claim API.
- Added indexes for user and draft-type lookups and expired-draft processing.
- Added a claim-specific unique index to prevent more than one claim draft for the same user.
- Added reusable draft-store entity, repository and service classes under the `draftstore` package.
- Added `DraftType` to provide draft-type identifiers and retention periods at runtime.
- Added a claim-specific service to handle claim draft behaviour and select the `DRAFT_CLAIM` type.
- Added service methods to create, retrieve, update and delete drafts.
- Set claim draft expiry to 180 days from the initial `draftClaimCreatedAt`.
- Kept `draftClaimCreatedAt`, `createdAt` and `expiresAt` unchanged when drafts are updated.
- Returned the existing active draft when creation is requested more than once for the same user.
- Added removal and replacement of an expired claim draft when a new draft is created.
- Excluded expired drafts from retrieval, active-draft lookup and updates.
- Added authenticated `POST`, `GET`, `PUT` and `DELETE` endpoints under `/dashboard/draft-claims`.
- Used the authenticated bearer token to identify the draft owner and prevent cross-user access.
- Added `404 Not Found` handling for missing, expired or unauthorised draft claims.
- Added request validation and defensive copying of draft payloads.
- Updated test names to consistently use the agreed `should` and `when` convention.
- Added unit and controller tests for generic draft storage and claim-specific behaviour.
- Added PostgreSQL integration tests for persistence, recovery, duplicate prevention, repeated creation, fixed expiry, expired-draft replacement, cross-user access and deletion.
- Improved integration-test isolation and cleaned up the draft-store production and test code.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/civil/controllers/DraftClaimControllerIntegrationTest.java
```diff
@@ -0,0 +1,372 @@
...
+
+        DraftStoreEntity draftInDb = draftStoreRepository.findById(createdDraftId)
+            .orElseThrow(() -> new AssertionError("Draft claim should be persisted in database"));
+
+        assertThat(draftInDb.getPayload()).extracting("step").isEqualTo("claimant-details");
+        assertThat(draftInDb.getUserId()).isEqualTo(USER_ID);
+    }
+
+    @Test
+    void repeatedCreationReturnsExistingDraftWithoutChangingIt() throws Exception {
```
toyin-hmcts: same comment for a few of the tests in this file, I think stick to a single naming convention i.e. "should...when" as you did above. This will make it clear how the test/system is expected to behave and what is being checked.

src/main/resources/db/migration/V2026_07_06_1417__DTSCCI-5548-create-draft_store_tables.sql
```diff
@@ -0,0 +1,32 @@
...
+VALUES (1, 'DRAFT_CLAIM');
+
+CREATE TABLE dbs.draft_store
+(
+  id UUID NOT NULL,
+  user_id VARCHAR(200) NOT NULL,
+  case_id VARCHAR(200),
+  draft_type_id INT NOT NULL,
+  payload JSONB NOT NULL,
+  draft_claim_created_at TIMESTAMP NOT NULL,
```
toyin-hmcts: Don't quite understand the purpose of this column in the table? What's the difference between this and created_at. Furthermore, for other draft types like draft_responses, this column would have to be null or it doesn't make sense. Feels to specific to draft claims.

src/main/java/uk/gov/hmcts/reform/dashboard/services/DraftStoreService.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+
+@Service
+@Transactional
+@Slf4j
+public class DraftStoreService {
+
+    private static final int DRAFT_CLAIM_TYPE_ID = 1;
```
toyin-hmcts: doesn't feel like the right choice to hard code the draft_type here since this service is for the whole draft_store rather than just for draft_claims. 
This class also feels misplaced for the same reason, it wouldn't just be for the dashboard since any page using drafts will call this service.

src/main/java/uk/gov/hmcts/reform/dashboard/services/DraftStoreService.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final long DRAFT_EXPIRY_DAYS = 180;
```
toyin-hmcts: I would abstract the calculation of this value more since there's a high chance that for different types of drafts, different expiry days will be required. so it needs to be calculated at runtime.

toyin-hmcts (changes requested):

src/integrationTest/java/uk/gov/hmcts/reform/civil/controllers/DraftClaimControllerIntegrationTest.java
AbuSalamHMCTS: Updated the tests in this file to consistently use the agreed should and when naming convention.

src/main/java/uk/gov/hmcts/reform/dashboard/services/DraftStoreService.java
AbuSalamHMCTS: Moved the generic entity, repository and service into a separate draftstore package. DraftStoreService now receives the draft type at runtime, while claim-specific behaviour and the DRAFT_CLAIM selection remain in DraftClaimService.

src/main/java/uk/gov/hmcts/reform/dashboard/services/DraftStoreService.java
AbuSalamHMCTS: Moved the retention period into DraftType. The generic service now calculates the expiry date using the supplied draft type, allowing future draft types to have their own retention periods.

src/main/resources/db/migration/V2026_07_06_1417__DTSCCI-5548-create-draft_store_tables.sql
AbuSalamHMCTS: The ticket wants both. It says to store draftClaimCreatedAt, expiresAt, and created and updated timestamps.

toyin-hmcts: @AbuSalamHMCTS  ahh okay sorry about the confusion. From my point of view, any draft only requires a create_at timestamps which can then drive the expiry logic. A specific field for draftClaimCreatedAt is not required. I think the original ticket didn't consider other types of drafts aside from draft_claims.

src/main/java/uk/gov/hmcts/reform/draftstore/DraftType.java
```diff
@@ -0,0 +1,20 @@
...
+@Getter
+@RequiredArgsConstructor
+public enum DraftType {
+
+    DRAFT_CLAIM(1, 180);
+
+    private final int id;
+    private final long retentionDays;
+
+    public OffsetDateTime calculateExpiry(OffsetDateTime draftCreatedAt) {
```
toyin-hmcts: does this logic handle the scenario where the draft type is not defined i.e. can we set a default retentionDays in cases where it has not been agreed  yet but we still want it to get deleted. I.e. 30 days.

toyin-hmcts (changes requested):

src/main/resources/db/migration/V2026_07_06_1417__DTSCCI-5548-create-draft_store_tables.sql
AbuSalamHMCTS: Resolved.

src/main/java/uk/gov/hmcts/reform/draftstore/DraftType.java
AbuSalamHMCTS: Resolved.

src/main/java/uk/gov/hmcts/reform/draftstore/DraftType.java
```diff
@@ -0,0 +1,36 @@
...
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+import java.time.OffsetDateTime;
+
+@Getter
+@RequiredArgsConstructor
+public enum DraftType {
+
```
gmikerlemos: Blocking: DTSCCI-5548 now requires draft claims to be recoverable for 180 days, but this configures `DRAFT_CLAIM` with a 30-day retention period. Please change this to 180 and update the corresponding tests. Otherwise the API will stop returning drafts after 30 days while CUI may tell users they are recoverable for 180 days.

src/main/java/uk/gov/hmcts/reform/draftstore/services/DraftStoreService.java
```diff
@@ -0,0 +1,128 @@
...
+            userId,
+            caseId,
+            draftType.getId(),
+            copyPayload(payload),
+            now,
+            now,
+            draftType.calculateExpiry(now)
+        );
+        log.info("Creating draft type={} draftId={}", draftType, draft.getId());
+        return draftStoreRepository.save(draft);
```
gmikerlemos: Non-blocking concurrency issue: `DraftClaimService.createDraftClaim` performs a read-before-create, while the database enforces one draft claim per user with a unique partial index. Two simultaneous first-time requests can both observe no draft and reach this save; one will then fail with a uniqueness violation and return a server error instead of loading the draft created by the competing request. Please consider handling the constraint violation by re-reading the active draft, or using an atomic/upsert approach, and add a concurrent-create test.

src/main/java/uk/gov/hmcts/reform/draftstore/DraftType.java
toyin-hmcts: Ticket updated accordingly.

toyin-hmcts (approved):

toyin-hmcts (approved):

PR #8136, opened 2026-07-16
DTSCCI-5875 Remove PII from civil-service logs

gmikerlemos:
### JIRA link (if applicable)

[DTSCCI-5875](https://tools.hmcts.net/jira/browse/DTSCCI-5875)

### Change description

- Removes `CaseData`, raw case JSON, `Party`, `Address`, and party-name values from audited logging statements.
- Retains operational identifiers, including case, user, task, document, notification, and reference IDs.
- Suppresses financial and payment details from audited logs.
- Masks email addresses retained for notification troubleshooting.
- Adds defence-in-depth Logback redaction for confirmed PII and financial values in log messages and exception output.
- Adds regression coverage proving operational identifiers are retained while sensitive values are redacted.

### Testing

- Full `./gradlew check`
- Focused redaction tests
- Checkstyle for main, test, integration-test, and contract-test sources

### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No


mapingo (approved):

anirudha-deshpande (approved):

ruban72 (commented):
Reviewed — solid, well-targeted PR. The statement-level fixes cover exactly the civil-service items from the audit, the exception logging is corrected properly (`log.error("… failed", caseId, e)` passes the throwable as the last arg), and the redactor tests are genuinely good (JSON, Lombok toString, retain-identifiers-vs-redact-financial, multi-line, null). The regex is carefully built — the `(?<![\w])` lookbehind correctly keeps `caseReference`/`ipAddress` while redacting `paymentReference`, and `(?i)` handles `postCode`/`postcode`.

One important point to resolve, then a few minors.

### Important: the redactor does not cover the App Insights path (verified)
civil-service ships logs to App Insights via the **Application Insights Java agent v3.7.8** (`Dockerfile` + `lib/applicationinsights.json`). The 3.x agent auto-instruments Logback and captures each log **event** (`event.getFormattedMessage()` + MDC), which bypasses the console appender's encoder/`PatternLayout`. The `%pii` converter is only wired into `CONSOLE_LOG_PATTERN`, so it redacts stdout/console output but **not** the `AppTraces` telemetry sent to App Insights — which is the sink where the PII was reported. `applicationinsights.json` also has no telemetry processors, so nothing else redacts that path.

The statement-level removals in this PR still protect App Insights (PII never enters the event), so the primary remediation is sound. But the defence-in-depth backstop has a gap for the AI-agent path: a future stray `log.info(party)` would be redacted on console yet still leak to App Insights.

Suggestion: add redaction on the App Insights path too, e.g. an AI-agent telemetry processor (`preview.processors`, an `attribute` mask on the log `message`) in `applicationinsights.json`, or event-level redaction. If we'd rather keep the redactor console-only, it's worth stating that explicitly and noting App Insights relies on the statement-level removals.

### Minor / verify
- **`logback.xml` vs Spring levels:** there was no `logback-spring.xml` before, so this new `logback.xml` will be picked up (good). Since it's `logback.xml` (not `-spring`), please confirm per-package `logging.level.*` from `application.yaml` still apply, and that `reform-base.xml` doesn't re-define `CONSOLE_LOG_PATTERN` after the `<include>` (which would drop the `%pii` wrapper). A quick check of a deployed log line covers both.
- **Value boundary:** the redaction value group stops at `,` `)` `}` `\r` `\n`, so a free-text value containing a comma leaks the part after it. Fine for structured/Lombok output, just noting.
- **Performance:** `redact()` runs two regexes (one ~45-alternative with lookbehind) on every log line incl. stack traces. It's on `ASYNC_CONSOLE` so off the request thread — acceptable, worth being aware of for high-volume ERROR logging.
- **Nit:** `maskEmail` masks every char except `@` (reveals length only) — safe, just more aggressive than the ticket's "keep first character".

Overall: happy with the code and the coverage. The only thing I'd want resolved before merge is confirming whether App Insights ingestion needs its own redaction (the point above), since that's the sink the ticket is really about.

PR #8161, opened 2026-07-21
DTSCCI-5905: Verify hearing notice attendee names against CCD

ruban72:

https://tools.hmcts.net/jira/browse/DTSCCI-5905

### Change description ###

Adds an automated, read-only way to verify that generated HMC hearing notices carry the correct party (attendee) names, so we do not have to eyeball 200+ cases by hand.

**Why**

Hearing notice party names are rendered from CCD CaseData at generation time (`HearingNoticeHmcGenerator`). With CCD as the source of truth, a notice is only incorrect if it is *stale* - generated before a name was corrected in CCD. This task detects those stale notices.

**What**

- `VerifyHearingNoticeNamesTask` - a read-only migration task. For each case it builds the expected names exactly as the generator does (minor -> litigation friend, otherwise party name; applicant1/2 + respondent1/2), downloads the notice PDFs from CDAM (`getHearingDocuments()` / `getHearingDocumentsWelsh()`) via the existing `DocumentDownloadService`, extracts the text with PDFBox, and logs one structured line per document: `VERIFY_HEARING_NOTICE case=.. document=.. result=MATCH|STALE|NO_NOTICE|ERROR missingNames=..`. Results are read back from App Insights. The task returns CaseData unchanged.

- Read-only support in the migration framework so verification tasks do **not** write to case history:
  - `MigrationTask.isReadOnly()` (default `false`) is the opt-in.
  - `AsyncCaseMigrationService` fetches the case via `CoreCaseDataService.getCase(..)` (a plain read - no state constraint, case-type agnostic) and skips `startUpdate`/`submitUpdate` entirely, so no `UPDATE_CASE_DATA` event is recorded.

**Tests**

- `VerifyHearingNoticeNamesTaskTest` (match, stale, no-notice, download failure, null reference) - builds real in-memory PDFs with PDFBox.
- `AsyncCaseMigrationServiceTest` - 2 new tests asserting the read-only path calls `getCase` and never calls start/submit (civil + GA).

The mutating migration path is unchanged; the new behaviour only triggers when a task opts in via `isReadOnly()`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/civil/handler/migration/VerifyHearingNoticeNamesTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/civil/handler/migration/VerifyHearingNoticeNamesTask.java
@@ -1,0 +1,182 @@
+package uk.gov.hmcts.reform.civil.handler.migration;
+
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pdfbox.Loader;
+import org.apache.pdfbox.pdmodel.PDDocument;
+import org.apache.pdfbox.text.PDFTextStripper;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.civil.bulkupdate.csv.CaseReference;
+import uk.gov.hmcts.reform.civil.config.SystemUpdateUserConfiguration;
+import uk.gov.hmcts.reform.civil.documentmanagement.model.CaseDocument;
+import uk.gov.hmcts.reform.civil.model.CaseData;
+import uk.gov.hmcts.reform.civil.model.Party;
+import uk.gov.hmcts.reform.civil.model.common.Element;
+import uk.gov.hmcts.reform.civil.service.UserService;
+import uk.gov.hmcts.reform.civil.service.documentmanagement.DocumentDownloadService;
+import uk.gov.hmcts.reform.civil.utils.PartyUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Read-only reporting migration task.
+ *
+ * <p>Hearing notice party names are rendered from CCD CaseData at generation time
+ * ({@code HearingNoticeHmcGenerator}), so with CCD as the source of truth a notice is
+ * correct if its printed names still match the current case. This task downloads each
+ * HMC hearing notice PDF from CDAM and checks that the current CCD claimant/defendant
+ * names appear in the document text. A notice whose text no longer contains a current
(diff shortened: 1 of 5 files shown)


mounikahmcts (approved):

src/main/java/uk/gov/hmcts/reform/civil/handler/migration/VerifyHearingNoticeNamesTask.java
```diff
@@ -0,0 +1,186 @@
...
+        List<CaseDocument> notices = hearingNoticeDocuments(caseData);
+
+        if (notices.isEmpty()) {
+            log.info("{} case={} result=NO_NOTICE expectedNames={}", LOG_PREFIX, caseId, expectedNames);
+            return caseData;
+        }
+
+        String authorisation = getSystemUserToken();
+
+        for (CaseDocument notice : notices) {
```
gmikerlemos, line 96: Could token acquisition be covered by the structured error handling? If `getSystemUserToken()` fails here, the outer migration service only emits its generic error, so this case produces no `VERIFY_HEARING_NOTICE ... result=ERROR` record and may disappear from the App Insights verification report. Moving this into a guarded block, or emitting a case-level structured `ERROR`, would keep the report complete.

src/main/java/uk/gov/hmcts/reform/civil/handler/migration/VerifyHearingNoticeNamesTask.java
```diff
@@ -0,0 +1,186 @@
...
+                if (!noticeText.contains(normalise(name))) {
+                    missing.add(name);
+                }
+            }
+
+            String result = missing.isEmpty() ? "MATCH" : "STALE";
+            log.info("{} case={} document=\"{}\" result={} expectedNames={} missingNames={}",
+                     LOG_PREFIX, caseId, fileName, result, expectedNames, missing);
+        } catch (Exception e) {
+            log.error("{} case={} document=\"{}\" result=ERROR message={}",
```
gmikerlemos, line 122: Could we avoid logging `expectedNames` and `missingNames` here? These values contain claimant, defendant, and potentially litigation-friend names, which would place personal data into centrally retained/searchable Application Insights logs. Logging stable party identifiers instead (for example, `APPLICANT_1` or `RESPONDENT_2`) would preserve enough information to identify which names need investigation without recording the names themselves.

PR #8169, opened 2026-07-22
DTSCCI-0000 Migration batch size should be configurable

mounikahmcts:


mapingo (approved):

gajanan-deshpande (approved):

src/main/resources/application.yaml
```diff
@@ -268,7 +268,7 @@ dashboard:
     url: ${CIVIL_SERVICE_URL:http://localhost:4000}
 
 migration:
-  batchsize: 100
+  batchsize: ${MIGRATION_BATCH_SIZE:100}
```
sabahirfan, line 271: you can still configure this using `MIGRATION_BATCHSIZE` or `MIGRATION_BATCH_SIZE` without this change. A magic by spring!

Spring's property name relaxation (which treats hyphens and underscores as interchangeable).  However, `MIGRATION_BATCHSIZE` is the correct and guaranteed approach

mounikahmcts: ah true, then i dont need this PR

